PR #15, opened 2015-02-13
eliminate the pointless empty guardedHandlers property

dherman:


mikesherov:
back out the gaurdedHandlers change until we can get conformation from @ariya here.

mikesherov:
because the rest you can land :+1:

mikesherov:
The directive change was to simply add a "directive" property to expression statement, so I think you might just accidentally botched this PR. Closing for now.

dherman:
Sorry I was sloppy. I've separated the guardedHandlers issue into issue #24 and I will submit a proper PR for the directive property.

Issue #5, opened 2015-02-12
Concise methods node


nzakas:
Esprima currently represents concise methods as `Property` whose `value` is a `FunctionExpression`. This is pretty confusing because ES5 style properties with function values are represented the same way. Even though the `method` flag is `true`, that still means the `FunctionExpression` node represents different syntax in each situation, and that can cause errors such as https://github.com/eslint/eslint/issues/1677

michaelficarra:
But getters and setters are also represented with a FunctionExpression node. It is only appropriate to re-use it yet again for methods.

nzakas:
Good point, and similar problem. Maybe a new property on `FunctionExpression` to make this distinction? As an AST consumer, having `FunctionExpression` representing different syntax depending on the parent makes things pretty confusing.

michaelficarra:
Tell me about it. But that problem exists all over the SpiderMonkey AST. `Identifier` and `BlockStatement` are possibly the worst offenders.

impinball:
@michaelficarra `Identifier`, in my experience, is only really for one purpose, to represent an identifier, whether used for an unquoted property name, variable, or function name. In my opinion, it just makes perfect sense. The problem with `BlockStatement` is that it's somewhat inconsistently used, and it's a little late to change it now, since tooling has largely gotten used to it being stable.

sebmck:
@impinball We have the opportunity to change things like that with ESTree.

RReverser:
@sebmck We should not change pre-ES6 types cardinally though, as I don't think we want to break all the existing tools.

sebmck:
@RReverser Sure, I agree, I just don't think these types of things should be taken off the table for consideration.

michaelficarra:
@sebmck: That's a slippery slope. If you're going to make pre-ES6 breaking changes, there's not much argument against just adopting Shift. We shouldn't be considering breaking changes like those in this repo.

impinball:
I like the original proposal in this bug, a new boolean `.method` property for ES6 methods. I highly doubt that someone is already checking `method: false` everywhere, and in tests, it's relatively simple (albeit repetitive) to fix.

@sebmck I was merely clarifying a couple specific points on why certain nodes were used, and I highly doubt any of that's going to change now (not to mention it is a little unintelligent to do so, since patching all the tools isn't exactly trivial for anything). I actually disagreed with one of them, and the other isn't something easily replaced.

[...]

PR #19, opened 2015-02-14
Add ES6 class definitions

sebmck:
Both acorn and esprima harmony follow this structure.

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -89,3 +89,52 @@ interface ArrayPattern <: Pattern {
     elements: [ Pattern | null ];
 }
 ```
+
+# Classes
+
+```js
+interface Class <: Node {
+    id: Identifier | null;
+    superClass: Expression;
+    body: ClassBody;
+}
+```
+
+## ClassBody
+
+```js
+interface ClassBody <: Node {
+    type: "ClassBody";
+    body: [ MethodDefinition ];
+}
+```
+
+## MethodDefinition
+
+```js
+interface MethodDefinition <: Node {
+    type: "MethodDefinition";
+    key: Identifier;
+    value: FunctionExpression;
+    kind: "" | "get" | "set";
+    computed: boolean;
+    static: boolean;
+}
+```
+
+## ClassDeclaration
+
+```js
+interface ClassDeclaration <: Class, Declaration {
+    type: "ClassDeclaration";
+    id: Identifier;
+}
+```
+
+## ClassExpression
+
+```js
+interface ClassExpression <: Class, Expression {
+    type: "ClassExpression";
+}
+```


mikesherov:
Can I get a :+1: from @ariya and @dherman confirming?

es6.md
```diff
@@ -89,3 +89,52 @@ interface ArrayPattern <: Pattern {
...
+```
+
+## MethodDefinition
+
+```js
+interface MethodDefinition <: Node {
+    type: "MethodDefinition";
+    key: Identifier;
+    value: FunctionExpression;
+    kind: "init" | "get" | "set";
```
RReverser: @sebmck Isn't it `""` for regular methods in current implementation? `init` seems to make sense for `Property` only (where you actually initialize it with some value), however I'm more than open to specifying `"method"` or smth here.

sebmck: Ah, yes it is. Copy+paste from `Property` fail.

es6.md
```diff
@@ -89,3 +89,52 @@ interface ArrayPattern <: Pattern {
...
+    body: ClassBody;
+}
+```
+
+## ClassBody
+
+```js
+interface ClassBody <: Node {
+    type: "ClassBody";
+    body: [ MethodDefinition ];
```
RReverser: How about renaming this to `methods` instead? For `BlockStatement`, we can't change it, but here I think we could fix the `classStmt.body[0].body[1].value` chain into smth more meaningful.

sebmck: Possibly. Just wanted to align this PR with the current pseudo-standardised behaviour.

es6.md
```diff
@@ -89,3 +89,52 @@ interface ArrayPattern <: Pattern {
...
+```
+
+## MethodDefinition
+
+```js
+interface MethodDefinition <: Node {
+    type: "MethodDefinition";
+    key: Identifier;
+    value: FunctionExpression;
+    kind: "" | "get" | "set";
```
RReverser: Since comment was removed - I'll update - how about using `"method"` or smth here instead of empty string?

es6.md
mikesherov: Hesitant to rename it considering it's already interoperable and this seems like a minor nit.

RReverser: Sure, thanks for that, just trying to bring some discussion before we get this as standardized behavior.

mikesherov: Yeah, should always discuss ahead of time. Glad you're reviewing these PRs ;-)

es6.md
mikesherov: perhaps file a followup issue for this? Reasoning is that Esprima is launching classes soon and also a feature to test downstream projects.

Once we gather data on how likely this is to break for consumers, we can better figure out the urgency and risk for changing it.

RReverser: @mikesherov Maybe. I thought we can either discuss it here before merging PR to spec, or might make sense to merge PR and then post an issue. Thoughts?

mikesherov: If we plan on having a "living standard", I think it's safe to merge anything that documents current interop and then propose breaking changes separately.

RReverser: Cool, let's do it then, so I could post an issue with reference to living spec.

fkling: Does this have any impact on #5 ? Afaik the spec uses the same production rule for methods in object expressions and class bodys. Do we want two different node types? (it may be necessary, just thinking out loud)

RReverser: Not really the same - class method definition is something that can appear and be used only on this class and instances, while in object expression it's just shorthand syntax for regular `Property` with `FunctionExpression` that can be replaced with any other `value` later.

fkling: They are [the same in the spec however](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-method-definitions), that's why I was wondering if it makes sense to treat them the same in an AST as well.

sebmck:
So is the consensus that this is good to merge since it documents the current pseudo-standardised behaviour of Acorn/Esprima and changes can be made and discussed in separate PRs?

mikesherov:
Yes.

dherman:
:+1:

sebmck:
What's the process going to be for merging pull requests and agreeing on changes?

mikesherov:
@sebmck, sounds like if there is already interop between acorn, SpiderMonkey, and Esprima, we should just document, and have 1 person from each give a plus one.

In case of proposed modifications, I'd imagine a discussion period on the Issue followed by plus one from each SpiderMonkey, Esprima, and acorn.

If we can't agree on a proposed modification, we should at least document the differences and attempt at a later date to resolve.

@sebmck, @dherman, sound ok? If so, we should document this process.

It's a bit suboptimal because it ignores other potential implementations, so I'm open to suggestions.

dherman:
I think that's a good starting point, and it's all on GH so it's open participation, but if people feel we are moving too fast we can adjust as we go.

sebmck:
@mikesherov Yeah, documenting it sounds good, there shouldn't be any ambiguity with these types of decisions.

PR #20, opened 2015-02-15
Allow only Pattern and MemberExpression in assignment target

RReverser:
If I don't miss anything, any other expression can't appear in left side of `AssignmentPattern`.

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -37,7 +37,7 @@ extend interface VariableDeclaration {
 
 ```js
 extend interface AssignmentExpression {
-    left: Pattern | Expression;
+    left: Pattern | MemberExpression;
 }
 ```
 


sebmck:
`Identifier`?

RReverser:
@sebmck It already implements `Pattern`: https://github.com/estree/estree/blob/master/spec.md#identifier

sebmck:
Ah, my bad.

michaelficarra:
Same thing with `ForInStatement` and `ForOfStatement`?

RReverser:
@michaelficarra See #21. Or do you mean something else?

RReverser:
@michaelficarra Ah, got it, you're right.

dherman:
Interesting question. Unless I'm misunderstanding the specs, *LeftHandSideExpression* has been very liberal all the way through ES5, even allowing function calls. Not until ES6 does the [*IsValidSimpleAssignmentTarget* static semantics](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-static-semantics-static-semantics-isvalidsimpleassignmenttarget) reject anything other than productions that produce `"MemberExpression"` nodes.

But ES6 is not describing what real engines do. If you try executing:
```js
for (console.log() in Object.create(null));
```
In Firefox and Chrome it succeeds; in Safari it seems to parse fine but dynamically throw a ReferenceError. (I believe at least in old versions of IE there were some APIs that returned the ECMAScript Reference Type so you could actually legally execute something like `someIEAPI() = 42;`, but I'm having trouble turning up any search results on this.) Note that the above example parses just fine with SpiderMonkey's `Reflect.parse` and produces a `.left` property of type `"CallExpression"`.

I can see a couple options:
* Keep the type more liberal to reflect what real JS engines do.
* Restrict the type to MemberExpression and re-specify it with the more liberal type in one of the extension specs.

dherman:
For historical interest: if I'm reading correctly, [this old SpiderMonkey bug](https://bugzilla.mozilla.org/show_bug.cgi?id=319391) suggests that old versions of IE would return references from `eval` so you could use that on the left-hand side of an assignment. I chatted with Brendan and he recalled that `document.all(...)` would also return a reference in old IE.

RReverser:
Interesting (and weird) behavior.... Do you think we should really leave things that way?

dherman:
I'm inclined to say leave it with the more general type Expression since that is legal by ES5 and happens in practice. We could add a note that some implementations may optionally choose to restrict it to MemberExpression.

kaycoinsUSA (commented):
Done

PR #21, opened 2015-02-15
Make ForOfStatement to derive from ForInStatement

RReverser:
They are pretty similar from parser/traverser perspective, so I don't see a reason for duplicating node structures.

Also allows only `Pattern` and `MemberExpression` as iteration assignment target instead of too generic `Expression` (thanks @michaelficarra).

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -15,11 +15,8 @@ extend interface Function {
 ## ForOfStatement
 
 ```js
-interface ForOfStatement <: Statement {
+interface ForOfStatement <: ForInStatement {
     type: "ForOfStatement";
-    left: VariableDeclaration |  Expression;
-    right: Expression;
-    body: Statement;
 }
 ```
 


impinball:
Agreed. Although I can't exactly merge this one.

RReverser:
@impinball Well, I have merge/commit access, but willing to send PRs so that others can see and discuss if needed.

impinball:
Okay. Probably should've already been aware...

RReverser:
@dherman @ariya What would you say?

mikesherov:
This seems like IDL change only??? Nothing changing in implementation? Excuse the noobish question here.

RReverser:
Yep, just a cleanup (as was with ArrowExpression).

mikesherov:
If no code changes are required, then :+1:

sebmck:
Yup, simplifying node constructs shouldn't be contentious.

RReverser:
Cool.

PR #22, opened 2015-02-15
Remove already inherited definitions from ArrowExpression

RReverser:
`params`, `defaults`, `rest` and `generator` already exist in `Function` that `ArrowExpression` inherits from, so there is little sense in redefining those again as something specific to `ArrowExpression`.


impinball:
I like the simplicity here.

RReverser:
This one seems to be simple and non-breaking, so I'm merging it.

PR #23, opened 2015-02-15
Add ES6 Property extensions

RReverser:
Describes current behavior of both Esprima and Acorn to indicate shorthand property, method shorthand syntax and computed property.

Also, in `ObjectPattern` regular `Property` type is reused but with custom restrictions.

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -82,7 +82,7 @@ A `yield` expression.
 ## ObjectPattern
 
 ```js
-interface PatternProperty <: Property {
+interface AssignmentProperty <: Property {
     type: "Property"; // inherited
     value: Pattern;
     kind: "init";
@@ -91,7 +91,7 @@ interface PatternProperty <: Property {
 
 interface ObjectPattern <: Pattern {
     type: "ObjectPattern";
-    properties: [ PatternProperty ];
+    properties: [ AssignmentProperty ];
 }
 ```
 
--- a/es6.md
+++ b/es6.md
@@ -39,6 +39,13 @@ extend interface VariableDeclaration {
 extend interface AssignmentExpression {
     left: Pattern | Expression;
 }
+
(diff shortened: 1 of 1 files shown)


es6.md
```diff
@@ -75,9 +82,16 @@ A `yield` expression.
 ## ObjectPattern
 
 ```js
+interface PatternProperty <: Property {
+    type: "Property"; // inherited
+    value: Pattern;
+    kind: "init";
+    method: false;
```
impinball: Like the overall idea, but WRT `.method` specifically in PatternProperty, couldn't it wait for #5 to be resolved first?

RReverser: @impinball This is something that is already present in implementations; since this is a living standard, nothing stops us from changing docs when implementations will be changed too.

es6.md
```diff
@@ -75,9 +82,16 @@ A `yield` expression.
 ## ObjectPattern
 
 ```js
+interface PatternProperty <: Property {
```
fkling: Any specific reason to use `PatternProperty` instead of `PropertyPattern` (which would be more consistent with other Pattern names)?

es6.md
impinball: Okay. Probably worth mentioning in that bug, then.

es6.md
RReverser: @fkling Named it so because in this case it's not "property pattern" as it's not pattern itself, but rather "pattern's property". However exact name does not matter here since it doesn't affect `type` property and is used only for internal needs.

fkling: True.

RReverser: @fkling I like it, updated.

RReverser:
Since this is an already implemented behavior, I'll just merge this for now.
If implementations change, we can reflect it via another PR.

PR #26, opened 2015-02-16
Add experimental documentation

### LINKED ISSUE #18: Tracking early proposals
Some early proposals will need to be parsed before they have been completely finalised for use in linters and most notably transpilers. To ensure interoperability an AST needs to be agreed upon prior to their finalisation.

How should these early proposals be regulated?

sebmck:
Closes #18.

Included are some ES7 Async Function definitions. I've decided to limit the scope to proposals that are Stage 0 or above. I'm not fussed with the stage minimum so if anyone has any problems with it then it can be raised.

### DIFF
--- a/experimental.md
+++ b/experimental.md
@@ -1,0 +1,23 @@
+This document specifies extensions to the core ESTree AST types to support ES proposals that are at least [stage 0 or above](https://github.com/tc39/ecma262).
+
+**NOTE:** This document may change at anytime and **should not** be seen as stable.
+
+# [Async Functions](https://github.com/lukehoban/ecmascript-asyncawait)
+
+## Function
+
+```js
+extend interface Function {
+    async: false;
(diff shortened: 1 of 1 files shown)


sebmck:
Going to merge this in a few days unless there are any objections/suggestions.

experimental.md
```diff
@@ -0,0 +1,21 @@
+This document specifies extensions to the core ESTree AST types to support ES proposals that are at least [stage 0 or above](https://github.com/tc39/ecma262). This document is subject to change and should not be seen as stable.
```
mikesherov: Can we have slightly stronger wording here? Something mentioning that "may change at anytime"?

sebmck: Sure. Updated, let me know what you think.

mikesherov:
:+1:

PR #29, opened 2015-02-16
Add RegexLiteral definition

### LINKED ISSUE #27: Standardize new RegExp literal AST format
In Esprima, `/foo/gi` now gives:

```js
{
  type: "Literal",
  value: /foo/gi,
  regex: {
    pattern: "foo",
    flags: "gi"
  }
}
```

The advantage of having the separate `regex` property is that even in environments that don’t support a given flag (e.g. the `u` flag) the regex can still be represented in the AST. The `value` property is `null` in such a case.

See <https://code.google.com/p/esprima/issues/detail?id=557#c4>.

cc @sebmck @michaelficarra

sebmck:
Closes #27. Both Esprima and Acorn follow this representation.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -603,6 +603,20 @@ interface Literal <: Node, Expression {
 
 A literal token. Note that a literal can be an expression.
 
+### RegexLiteral
+
+```js
+interface RegexLiteral <: Literal {
+  regex: {
+    pattern: string;
+    flags: string;
+  };
+}
+```
+
+The `regex` property allows regexes to be represented in environments that don’t
+support certain flags such as `y` or `u`. In environments that don't support
+these flags `value` will be `null` as the regex can't be represented natively.
 
 ## UnaryOperator
 


spec.md
```diff
@@ -603,6 +603,19 @@ interface Literal <: Node, Expression {
...
+```js
+interface RegexLiteral <: Literal {
+  regex: {
+    pattern: string;
+    flags: string;
+  };
+}
+```
+
+The `regex` property allows regexes to be represented in environments that don’t
```
mathiasbynens: Should there be a note saying that the `value` property can be `null` in such cases?

RReverser: Yeah, I guess so.

impinball: This is really helpful.

sebmck: Oh yeah, knew I forgot something.

sebmck: Updated.

RReverser:
I guess we can merge this as it reflects already working implementation. Thoughts?

sebmck:
Merging this since it reflects the behaviour of Espree, Esprima and Acorn and there don't appear to be any objections.

mikesherov:
Yeah, just as a note. Consider anything that is in all 3 as "merge at will". Thanks for moving this forward. :)

Issue #11, opened 2015-02-13
Import Specifiers: slightly inconsistent between Acorn and Esprima


impinball:
Currently, here's the output of each import declaration type (in pseudo-CSON for ease of typing). I also have relevant bugs within this repo about each, if I can find one.

Default import (`import foo from 'bar'`):

```coffee
# Acorn
type: 'ImportSpecifier'
id: Identifier('foo')
name: null
default: true

# Esprima
type: 'ImportDefaultSpecifier'
id: Identifier('foo')

# SpiderMonkey
type: 'ImportSpecifier'
id: Identifier('default')
name: Identifier('foo')
```

Named import (`import {foo} from 'bar'`):

```coffee
# Acorn
type: 'ImportSpecifier'
id: Identifier('foo')
name: null
default: false

# Esprima
type: 'ImportSpecifier'
id: Identifier('foo')
name: null

# SpiderMonkey
type: 'ImportSpecifier'
id: Identifier('foo')
name: Identifier('foo')
```

Namespace import (`import * as foo from 'bar'`):
Bugs: #9

```coffee
# Acorn
type: 'ImportBatchSpecifier'
name: Identifier('foo')

# Esprima
type: 'ImportNamespaceSpecifier'
id: Identifier('foo')

# SpiderMonkey
# (unsupported)
```

Aliased import (`import {foo as bar} from 'bar'`):

```coffee
# Acorn
type: 'ImportSpecifier'
id: Identifier('foo')
name: Identifier('bar')
default: false

# Esprima
type: 'ImportSpecifier'
id: Identifier('foo')
name: Identifier('bar')

# SpiderMonkey
type: 'ImportSpecifier'
id: Identifier('foo')
name: Identifier('bar')
```

I do wonder if the following would be better, a common interface for all of the import specifier types. There are a few backwards-compatibility breaks, but each of them are in areas where Acorn and Esprima already deviate. They are also noted explicitly inline. Any others are truthy/falsy compatible (just as Acorn and Esprima generally are).

```idl
// This is a virtual superclass, purely for clarification.
// No .default field, since it's generally redundant.
interface ImportSpecifierNode <| Node {
  type: string;
  id: Identifier;
  name: Identifier;
}

interface ImportDefaultSpecifier <| ImportSpecifierNode {
  type: "ImportDefaultSpecifier"; // Easier to test, consistent with Esprima
  id: Identifier;
  name: null;
}

interface ImportSpecifier <| ImportSpecifierNode {
[...]

PR #35, opened 2015-02-22
Add import nodes.

impinball:
This is to address the import side of things WRT #11 and friends.

Now, WRT ImportDeclaration:

This is the only difference between Acorn/SpiderMonkey and Esprima for ImportDeclarations. I'm personally biased towards using `Literal`, considering I don't otherwise get why it couldn't just be one.

[Acorn](https://github.com/marijnh/acorn/blob/master/acorn.js#L2745-L2759) and [SpiderMonkey](https://hg.mozilla.org/mozilla-central/file/86d2bb8bb1c9/js/src/jsreflect.cpp#l2141) both use `Literal` for node types, while [Esprima](https://github.com/ariya/esprima/blob/harmony/esprima.js#L3635-L3645) uses `ModuleSpecifier`.

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -150,7 +150,7 @@ interface ClassExpression <: Class, Expression {
 
 ## ImportDeclaration
 
-**FIXME:** Esprima's behavior differs from Acorn's and SpiderMonkey's behavior in that `source` is of type `ModuleSpecifier`, but is otherwise a `Literal`.
+**FIXME:** Esprima's behavior differs from Acorn's and SpiderMonkey's behavior in that `source` is of type `ModuleSpecifier`, which is otherwise identical to a `Literal`. See [Esprima bug 1077](https://github.com/jquery/esprima/issues/1077).
 
 ```js
 interface ImportDeclaration <: Declaration {
--- a/es6.md
+++ b/es6.md
@@ -170,13 +170,12 @@ interface ImportSpecifier {
 }
 ```
 
-An imported variable binding, e.g., `{foo}` in `import {foo} from "mod"` or `{foo as bar}` in `import {foo as bar} from "mod"`. The `imported` field represents the imported name from the module. The `local` field represents the bindings imported into the module's global scope. If it is a basic named import, such as in `import {foo} from "mod"`, both `imported` and `local` are equivalent `Identifier` nodes; in this case an `Identifier` node representing `foo`. If it is an aliased import, such as in `import {foo as bar} from "mod"`, the `imported` field is an `Identifier` node representing `foo`, and the `local` field is an `Identifier` node representing `bar`.
+An imported variable binding, e.g., `{foo}` in `import {foo} from "mod"` or `{foo as bar}` in `import {foo as bar} from "mod"`. The `imported` field refers to the name of the export imported from the module. The `local` field refers to the binding imported into the local module scope. If it is a basic named import, such as in `import {foo} from "mod"`, both `imported` and `local` are equivalent `Identifier` nodes; in this case an `Identifier` node representing `foo`. If it is an aliased import, such as in `import {foo as bar} from "mod"`, the `imported` field is an `Identifier` node representing `foo`, and the `local` field is an `Identifier` node representing `bar`.
 
 ## ImportDefaultSpecifier
 
 ```js
 interface ImportDefaultSpecifier {
-    imported: null;
     local: Identifier;
 }
 ```
@@ -187,7 +186,6 @@ A default import specifier, e.g., `foo` in `import foo from "mod.js"`.
 
 ```js
 interface ImportNamespaceSpecifier {
-    imported: null;
     local: Identifier;
 }
 ```
--- a/es6.md
+++ b/es6.md
@@ -145,3 +145,51 @@ interface ClassExpression <: Class, Expression {
     type: "ClassExpression";
 }
 ```
+
+# Modules
+
+## ImportDeclaration
+
+**FIXME:** Esprima's behavior differs from Acorn's and SpiderMonkey's behavior in that `source` is of type `ModuleSpecifier`, but is otherwise a `Literal`.
+
+```js
+interface ImportDeclaration <: Declaration {
+    specifiers: [ ImportSpecifier ];
+    source: Literal; // Or ModuleSpecifier? (Esprima deviation)
+}
+```
+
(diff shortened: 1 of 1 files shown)


es6.md
```diff
@@ -145,3 +145,51 @@ interface ClassExpression <: Class, Expression {
...
+}
+```
+
+An imported variable binding, e.g., `{foo}` in `import {foo} from "mod"` or `{foo as bar}` in `import {foo as bar} from "mod"`. The `imported` field represents the imported name from the module. The `local` field represents the bindings imported into the module's global scope. If it is a basic named import, such as in `import {foo} from "mod"`, both `imported` and `local` are equivalent `Identifier` nodes; in this case an `Identifier` node representing `foo`. If it is an aliased import, such as in `import {foo as bar} from "mod"`, the `imported` field is an `Identifier` node representing `foo`, and the `local` field is an `Identifier` node representing `bar`.
+
+## ImportDefaultSpecifier
+
+```js
+interface ImportDefaultSpecifier {
+    imported: null;
```
jeffmo: I think someone mentioned it on the other thread -- but given that this will always be `null`, it seems unnecessary for it to be in the syntax node. Can we just kill this property?

es6.md
```diff
@@ -145,3 +145,51 @@ interface ClassExpression <: Class, Expression {
...
+}
+```
+
+A default import specifier, e.g., `foo` in `import foo from "mod.js"`.
+
+## ImportNamespaceSpecifier
+
+```js
+interface ImportNamespaceSpecifier {
+    imported: null;
```
jeffmo: (same)

es6.md
```diff
@@ -145,3 +145,51 @@ interface ClassExpression <: Class, Expression {
...
+## ImportSpecifier
+
+```js
+interface ImportSpecifier {
+    imported: Identifier;
+    local: Identifier;
+}
+```
+
+An imported variable binding, e.g., `{foo}` in `import {foo} from "mod"` or `{foo as bar}` in `import {foo as bar} from "mod"`. The `imported` field represents the imported name from the module. The `local` field represents the bindings imported into the module's global scope. If it is a basic named import, such as in `import {foo} from "mod"`, both `imported` and `local` are equivalent `Identifier` nodes; in this case an `Identifier` node representing `foo`. If it is an aliased import, such as in `import {foo as bar} from "mod"`, the `imported` field is an `Identifier` node representing `foo`, and the `local` field is an `Identifier` node representing `bar`.
```
jeffmo: > The `imported` field represents the imported name from the module.

Reads slightly confusing. Perhaps: "The `imported` field refers to the name of the export coming from the source module."

> The `local` field represents the bindings imported into the module's global scope

Let's say "the local module scope" rather than "global scope"

es6.md
impinball: Probably a good idea.

es6.md
impinball: WDYT about this?

jeffmo:
lgtm

es6.md
```diff
@@ -145,3 +145,49 @@ interface ClassExpression <: Class, Expression {
     type: "ClassExpression";
 }
 ```
+
+# Modules
+
+## ImportDeclaration
+
+**FIXME:** Esprima's behavior differs from Acorn's and SpiderMonkey's behavior in that `source` is of type `ModuleSpecifier`, but is otherwise a `Literal`.
```
sebmck: Is this necessary? The point of ESTree is to remove these discrepancies instead of documenting them.

impinball: That's why it says `FIXME`. It's something that needs fixed. I'm fine holding onto this until there can be some sort of conclusive agreement. Note that [there is a similar discrepancy over ArrowExpression](https://github.com/estree/estree/blob/master/es6.md#arrowexpression) already committed, with a similar tag. It's something that needs to be worked out.

I just filed an Esprima bug over this: jquery/esprima#1077

sebmck: Ah right, thanks for the clarification.

es6.md
```diff
@@ -145,3 +145,49 @@ interface ClassExpression <: Class, Expression {
...
+# Modules
+
+## ImportDeclaration
+
+**FIXME:** Esprima's behavior differs from Acorn's and SpiderMonkey's behavior in that `source` is of type `ModuleSpecifier`, which is otherwise identical to a `Literal`. See [Esprima bug 1077](https://github.com/jquery/esprima/issues/1077).
+
+```js
+interface ImportDeclaration <: Declaration {
+    specifiers: [ ImportSpecifier ];
+    source: Literal; // Or ModuleSpecifier? (Esprima deviation)
```
caridy: this is going to be a problem in the future once we introduce a module specifier that is not a literal, and this is suppose to be in ES2016, the front runner at this point is: `import {name} from this module;`, where `this module` is an especial module specifier that can be used to access meta information about the module and the loader. this is the main reason why I've chosen `ModuleSpecifier` for esprima.

jeffmo: If it takes a literal or something else, the spec would just change to `literal|ModuleSpecifier` though.

jeffmo: (Using a union here also leaves room for a third "source" type. Most importantly though, it means all of these kinds of sources don't have to share a single unified data structure to represent their various bits)

caridy: Spec references that part of the `ImportDeclaration` (and the `ExportDeclaration` as well) as a `ModuleSpecifier`, assuming it will evolve over time. Really, what I'm trying to convey here is the idea that after the `from`, we will alway have a module specifier that could be a literal, rather than saying after the `from` we will have a literal, or something else.

jeffmo: Right. Do you agree that, if in the future more forms of the source get added, we'd still be able to represent a structure that's additive to just `Literal` by unioning that/those form(s) in the spec here? And that that would still be backward compatible?

Or are you seeing a different issue that I'm missing?

mikesherov:
We resolved yesterday to go with Literal *for now* in Esprima to achieve interop. @jeffmo's suggestion of a future Union in the spec would be back compat friendly considering it's additive nature. 

@caridy we'd like to move forward here. Can we merge this and revisit in es2016 knowing that we can make a future friendly change later?

caridy:
:+1:

mikesherov:
Merging considering consensus has been achieved and Esprima has a PR for the new behavior.

PR #37, opened 2015-03-02
Replace defaults/rest with AssignmentPattern/RestElement.

### LINKED ISSUE #3: Acorn: AssignmentPattern/RestElement vs Esprima: defaults/rest
Currently acorn uses `AssignmentPattern` and `RestElement` while esprima uses a `defaults` array and `rest` property.

```javascript
function foo(bar = "", ...items) {}
```

**Esprima**

```javascript
{
  type: "FunctionDeclaration",
  params: [{ type: "Identifier", name: "bar" }],
  defaults: [{ type: "Literal", value: "" }],
  rest: { type: "Identifier", name: "items" },
  body: {
    type: "BlockStatement",
    body: []
  }
}
```

**Acorn**

```javascript
{
  type: "FunctionDeclaration",
  params: [
    {
      type: "AssignmentPattern",
      left: { type: "Identifier", name: "bar" },
      right: { type: "Literal", value: "" }
    },
    {
      type: "RestElement",
      argument: { type: "Identifier", name: "items" }
    }
  ],
  body: {
    type: "BlockStatement",
    body: []
  }
}
```

I had to adapt 6to5 to the acorn behaviour and I must say in my opinion it makes it far easy to work with.

 - https://github.com/marijnh/acorn/issues/182
 - https://bugzilla.mozilla.org/show_bug.cgi?id=1093504

benjamn:
Submitting this as a pull request so that there can be discussion about the interface definition itself, which is the primary concern of this repository.

Closes #3.

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -4,8 +4,6 @@ This document specifies the extensions to the core ESTree AST types to support t
 
 ```js
 extend interface Function {
-    defaults: [ Expression ];
-    rest: Identifier | null;
     generator: boolean;
 }
 ```
@@ -94,6 +92,23 @@ interface ObjectPattern <: Pattern {
 interface ArrayPattern <: Pattern {
     type: "ArrayPattern";
     elements: [ Pattern | null ];
+}
+```
+
+## RestElement
+
+```js
+interface RestElement <: Pattern {
+    argument: Pattern
+}
+```
+
+## AssignmentPattern
+
+```js
+interface AssignmentPattern <: Pattern {
+    left: Pattern,
+    right: Expression
 }
 ```
 
--- a/spec.md
+++ b/spec.md
@@ -50,8 +50,6 @@ A complete program source tree.
 interface Function <: Node {
     id: Identifier | null;
     params: [ Pattern ];
-    defaults: [];
-    rest: null;
     body: BlockStatement;
     generator: false;
     expression: false;


RReverser:
LGTM.

mikesherov:
@michaelficarra, @ariya can I get a :+1: here from Esprima. This is as discussed in the last Esprima meeting.

mikesherov:
Also, @dherman?

michaelficarra:
I still disagree with adding RestElement, but I can confirm that this is the interface that was agreed upon at the esprima meeting.

nzakas:
:+1:

RReverser:
Ping @dherman or should we merge this now?

mikesherov:
Well, Esprima team agreed to make change based on confirm from SM, so I'd like to wait for dherman.

RReverser:
> based on confirm from SM

I just thought that already existing confirmed bug in their Bugzilla (https://bugzilla.mozilla.org/show_bug.cgi?id=1093504) would be enough.

mikesherov:
@rreverser that bug looks NEW to me with no sign from Mozilla they agree. We're all set, but just want to ensure @dherman gives explicit signal. Forgive me for being a bit pedantic here.

RReverser:
That issue was branched from https://bugzilla.mozilla.org/show_bug.cgi?id=932080 and agreed with @jorendorff and @Swatinem from SpiderMonkey, but I definitely do understand your concerns. Let's wait for @dherman then.

mikesherov:
BTW, the spec is missing a definition for `Pattern`, which a bunch of nodes inherit from. @benjamn can you add that definition. @RReverser thoughts?

mikesherov:
So, this PR needs to define `Pattern`, and it needs to have descriptions of all node types added, and needs to also augment functionExpression, functionDeclaration, ArrowFunctionExpression to include `RestElement` and `AssignmentPattern` as valid parameter types.

benjamn:
I thought the same thing at first, but the type of `Function.params` is `[Pattern]` already, so those alternatives are allowed. Changing it to `[Pattern | AssignmentPattern | RestElement]` would be redundant, since `AssignmentPattern <: Pattern` and `RestElement <: Pattern`.

mikesherov:
@benjamn fair enough. Missed that! Just need a def for pattern, and the other things I mentioned.

RReverser:
> Just need a def for pattern

@mikesherov It's already defined in [spec.md](https://github.com/estree/estree/blob/master/spec.md#patterns)

mikesherov:
OK, i'm a moron.

RReverser:
Don't say so.

RReverser:
@benjamn @mikesherov Actually, just thought on that a bit more, and it's not quite that simple (Liskov principle always beats us :smile:).

We will need `AssignmentPattern` to derive from `Node` and not `Pattern` and explicitly use `Pattern | AssignmentPattern` where both are allowed (`ArrayPattern.elements`, `ObjectPattern.properties`, `Function.params`).

Otherwise, structure allows:

1. To nest `AssignmentPattern` into `AssignmentPattern.left`, which is obviously wrong.
2. To use `AssignmentPattern` as `VariableDeclarator.id`, which would be cool and would avoid the need for `VariableDeclarator.init` and, thus, `VariableDeclarator` type in general, but we can't break pre-ES6 things, so also wrong.
3. To use `AssignmentPattern` as `CatchClause.param` which is also incorrect (?).
4. Use it in `AssignmentExpression.left`.

benjamn:
There are other `Pattern` subtypes that aren't forbidden by the AST type system, but that are forbidden by the spec. For instance, `PropertyPattern` and `RestElement` can't really appear on the left side of a default parameter binding, but I think we can agree they ought to be `<: Pattern`. In other words, I think it's fine to let `AssignmentPattern` remain `<: Pattern`.

Reusing `AssignmentPattern` for variable declarators seems weird, because the `.right` expression is not a default value in that case.

mikesherov:
OK, we're ready to merge this in as `RestElement` has landed in esprima 2.1, and `AssignmentPattern` will land in 3.0. @RReverser can we merge despite your late objection?

RReverser:
Of course. Those adjustions are not critical and non-breaking and can be easily introduced in future separately if needed.

mikesherov:
Hooray!

mikesherov:
@dherman just FYI we merged this.

RReverser:
Just a side note:

> Reusing AssignmentPattern for variable declarators seems weird, because the .right expression is not a default value in that case.

@benjamn It kind of is, because it's initial "default" value that can be changed later to another one. Anyway, as I said, it's not the case since such a change would be non-backward-compatible.

PR #38, opened 2015-03-02
Add export nodes - follow up of PR #35

caridy:
Related to #11, #35 and friends.

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -146,6 +146,54 @@ interface ClassExpression <: Class, Expression {
 }
 ```
 
+# Modules
+
+## ImportDeclaration
+
+**FIXME:** Esprima's behavior differs from Acorn's and SpiderMonkey's behavior in that `source` is of type `ModuleSpecifier`, but is otherwise a `Literal`.
+
+```js
+interface ImportDeclaration <: Declaration {
+    specifiers: [ ImportSpecifier ];
+    source: Literal; // Or ModuleSpecifier? (Esprima deviation)
+}
+```
+
+An import declaration, e.g., `import foo from "mod";`.
+
+## ImportSpecifier
+
+```js
+interface ImportSpecifier {
+    imported: Identifier;
+    local: Identifier;
+}
+```
+
+An imported variable binding, e.g., `{foo}` in `import {foo} from "mod"` or `{foo as bar}` in `import {foo as bar} from "mod"`. The `imported` field represents the imported name from the module. The `local` field represents the bindings imported into the module's global scope. If it is a basic named import, such as in `import {foo} from "mod"`, both `imported` and `local` are equivalent `Identifier` nodes; in this case an `Identifier` node representing `foo`. If it is an aliased import, such as in `import {foo as bar} from "mod"`, the `imported` field is an `Identifier` node representing `foo`, and the `local` field is an `Identifier` node representing `bar`.
+
+## ImportDefaultSpecifier
+
+```js
+interface ImportDefaultSpecifier {
+    imported: null;
+    local: Identifier;
+}
+```
+
+A default import specifier, e.g., `foo` in `import foo from "mod.js"`.
+
+## ImportNamespaceSpecifier
+
+```js
+interface ImportNamespaceSpecifier {
+    imported: null;
+    local: Identifier;
+}
+```
+
+A namespace import specifier, e.g., `* as foo` in `import * as foo from "mod.js"`.
+
 ## ExportDeclaration
 
 ```js
--- a/es6.md
+++ b/es6.md
@@ -225,10 +225,10 @@ interface ExportDefaultDeclaration <: Node {
 
 An export default declaration, e.g., `export default function () {};` or `export default 1;`.
 
-## ExportBatchDeclaration
+## ExportAllDeclaration
 
 ```js
-interface ExportBatchDeclaration <: Node {
+interface ExportAllDeclaration <: Node {
     source: Literal;
 }
 ```
--- a/es6.md
+++ b/es6.md
@@ -190,18 +190,19 @@ interface ImportNamespaceSpecifier {
 
 A namespace import specifier, e.g., `* as foo` in `import * as foo from "mod.js"`.
 
-## ExportDeclaration
+## ExportNamedDeclaration
 
 ```js
-interface ExportDeclaration <: Node {
-    default: boolean;
-    declaration: Declaration | Expression | null;
-    specifiers: [ ExportSpecifier | ExportBatchSpecifier ];
+interface ExportNamedDeclaration <: Node {
+    declaration: Declaration | null;
+    specifiers: [ ExportSpecifier ];
     source: Literal | null;
 }
 ```
 
-An export declaration, e.g., `export {foo} from "mod";` or `export default foo;`.
+An export named declaration, e.g., `export {foo, bar};`, `export {foo} from "mod";` or `export var foo = 1;`.
+
+_Note: Having `declaration` populated with non-empty `specifiers` or non-null `source` results in an invalid state._
 
 ## ExportSpecifier
 
@@ -214,10 +215,22 @@ interface ExportSpecifier {
 
 An exported variable binding, e.g., `{foo}` in `export {foo}` or `{bar as foo}` in `export {bar as foo}`. The `exported` field refers to the name exported in the module. The `local` field refers to the binding into the local module scope. If it is a basic named export, such as in `export {foo}`, both `exported` and `local` are equivalent `Identifier` nodes; in this case an `Identifier` node representing `foo`. If it is an aliased export, such as in `export {bar as foo}`, the `exported` field is an `Identifier` node representing `foo`, and the `local` field is an `Identifier` node representing `bar`.
 
-## ExportBatchSpecifier
+## ExportDefaultDeclaration
 
 ```js
-interface ExportBatchSpecifier {}
+interface ExportDefaultDeclaration <: Node {
+    declaration: Declaration | Expression;
+}
 ```
 
-A batch export specifier, e.g., `*` in `export * from "mod"`.
+An export default declaration, e.g., `export default function () {};` or `export default 1;`.
+
+## ExportBatchDeclaration
+
+```js
+interface ExportBatchDeclaration <: Node {
+    source: Literal;
+}
+```
+
+An export batch declaration, e.g., `export * from "mod";`.
--- a/es6.md
+++ b/es6.md
@@ -150,7 +150,7 @@ interface ClassExpression <: Class, Expression {
 
 ## ImportDeclaration
 
-**FIXME:** Esprima's behavior differs from Acorn's and SpiderMonkey's behavior in that `source` is of type `ModuleSpecifier`, but is otherwise a `Literal`.
+**FIXME:** Esprima's behavior differs from Acorn's and SpiderMonkey's behavior in that `source` is of type `ModuleSpecifier`, which is otherwise identical to a `Literal`. See [Esprima bug 1077](https://github.com/jquery/esprima/issues/1077).
 
 ```js
 interface ImportDeclaration <: Declaration {
--- a/es6.md
+++ b/es6.md
@@ -145,3 +145,35 @@ interface ClassExpression <: Class, Expression {
     type: "ClassExpression";
 }
 ```
+
+## ExportDeclaration
+
+```js
+interface ExportDeclaration <: Node {
+    default: boolean;
+    declaration: VariableDeclaration | Declaration | Expression | null;
+    specifiers: [ ExportSpecifier | ExportBatchSpecifier ];
+    source: Literal | null;
+}
+```
+
+An export declaration, e.g., `export {foo} from "mod";` or `export default foo;`.
+
+## ExportSpecifier
+
+```js
+interface ExportSpecifier {
+    exported: Identifier;
+    local: Identifier;
+}
+```
+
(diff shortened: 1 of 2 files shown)


es6.md
```diff
@@ -145,3 +155,37 @@ interface ClassExpression <: Class, Expression {
...
 ```
+
+# Modules
+
+## ExportDeclaration
+
+```js
+interface ExportDeclaration <: Declaration {
+    default: boolean;
+    declaration: VariableStatement | Declaration | HoistableDeclaration | null;
```
caridy: we might be missing something here.

mikesherov:
@caridy, are these the non controversial bits? I can't see anything that's not interop here?

If so, would love a :+1: from @sebmck, @dherman, and/or @rreverser

sebmck:
:+1:

mikesherov:
Maybe @jeffmo too? :+1: please!

RReverser:
How do we represent `export default (1+2)` here? I don't see where `Expression` goes.

mikesherov:
@RReverser well, that's embarrassing :-) You also just uncovered a bug in esprima.

es6.md
mikesherov: Where does `VariableStatement` come from?

sebmck: @mikesherov Looks like it's from the ES6 grammar, should be `VariableDeclaration` though obviously.

mikesherov: Also, as @RReverser notes, we need to support the ES6 definition of an `assignmentExpression` which includes a `conditionalExpression` which includes all binary expressions, like `export default (1 + 42);`

mikesherov: @sebmck @caridy I forgot where we discussed adding the HoistableDeclaration node. Did @michaelficarra or @ariya have issues there? I forgot...

es6.md
```diff
@@ -30,6 +30,16 @@ extend interface VariableDeclaration {
 }
 ```
 
+## HoistableDeclaration
```
mikesherov: @sebmck @caridy I forgot where we discussed adding the HoistableDeclaration node. Did @michaelficarra or @ariya have issues there? I forgot...

RReverser:
@mikesherov That's why we all are here :)

RReverser:
Probably that was a source for our disagreement on `HoistableDeclaration`.

I still don't see a reason for introducing new type here since, as soon as we add `Expression`, anonymous functions and classes get automatically covered by `Expression` node while named ones fall into `FunctionDeclaration` and `ClassDeclaration`.

Instead, if we add `HoistableDeclaration`, then we need to support another type with own `kind`s or something for distinguishing between `function` and `class` (as both are allowed in `export default`), support for non-named declarations, which will break assumptions in a lot of AST manipulation code I've seen and might bring more harm than benefits.

es6.md
```diff
@@ -145,3 +155,37 @@ interface ClassExpression <: Class, Expression {
     type: "ClassExpression";
 }
 ```
+
+# Modules
+
+## ExportDeclaration
+
+```js
+interface ExportDeclaration <: Declaration {
```
mikesherov: If we avoid making this and import extend from declaration, then not only are we following the specs definition of declaration, but the types accepted for the declaration field just says `declaration`

sebmck:
You're trying to force context dependent nodes into the SpiderMonkey (ESTree now I guess?) AST which isn't consistent with how the rest of the AST is specced. The beauty of not having a context-dependent AST is that I can manipulate it and shove nodes wherever I want and duplicate them without having to manipulate them and check for semantic differences such as different node types in different contexts like `HoistableDeclaration`.

RReverser:
> You're trying to force context dependent nodes into the SpiderMonkey

@sebmck Trying to figure out if that was addressed to me or someone else.

sebmck:
@RReverser Directed at the inclusion of an additional `HoistableDeclaration` node.

RReverser:
@sebmck :+1:

es6.md
mikesherov: Also note https://github.com/estree/estree/pull/39

jeffmo:
I think a wrapper-node for something like HoistedDeclaration would be overkill. I am in favor of making the FunctionDeclaration node's 'id' field nullable.

I realize the latter was controversial before, but I'd like to understand if there's someone who feels strongly different about the former (and why).

@rreverser: in our previous discussions about making FunctionDeclaration ids nullable (as they are in FunctionExpressions), I don't think we ever really got to the bottom of concretely describing what kinds of painful circumstances that would cause in real use cases. Having looked over most of our code with this question in mind, it would require almost zero changes to our code (which was a little surprising since I at least expected to have to change an expected type name somewhere in the ES6 module transpiration code -- but it worked out that I actually wasn't depending on that). It's a little hard to imagine any scenarios outside of any kind of module analysis code being affected by this to any significant degree since there are (currently) no other places in the grammar where an id field is optional on a func decl.

The important point I'd like to make, though, is that the spec now includes a form of function declaration that /is/ nameless though. This means (but doesn't guarantee) that at some point in the future the number of positions where this is possible may very well grow -- there's nothing that says nameless function declarations must forever going forward be relegated to the export position. In the scenario where they are extended to other parts of the grammar in the future, we're likely to have to deal with this fact anyway then...and it woils be even more awkward if, after making the breaking change elsewhere, we had a grandfathered-in version of export that uses FunctionExpression for no then-discernible reason except "historical convenience".

I say we "bite the bullet" (and I'm pretty confident it's more of a pleasantly soft and chewy bullet than one might think) and make this change now rather than deviating from the structure as described in the spec today only for it to be an odd gotcha scenario in the future when the historical context has faded.

caridy:
Ok, I have updated few things based on the feedback.

As for the `HoistedDeclaration`, @jeffmo suggestion makes a lot of sense, and it should be very straight forward, I have updated this PR to reflect that by removing `HoistedDeclaration` in favor of a function declaration with nullable id. Do we need the same for `export default class {}`?

es6.md
```diff
@@ -145,3 +145,35 @@ interface ClassExpression <: Class, Expression {
     type: "ClassExpression";
 }
 ```
+
+## ExportDeclaration
+
+```js
+interface ExportDeclaration <: Node {
+    default: boolean;
+    declaration: VariableDeclaration | Declaration | Expression | null;
```
mikesherov: Can remove `VariableDeclaration` as that's a declaration :-). Would also need #39 to land for this to be accurate, so that should be blocking this.

mikesherov:
Can someone point me to where HoistableDeclaration was discussed prior to this?

mikesherov:
@caridy can you rebase please?

caridy:
@mikesherov previous discussion about hoistable declaration is here: https://github.com/jquery/esprima/issues/311

mikesherov:
OK, so seems everyone is onboard with no hoistableDeclaration, including jquery/esprima folks in that discussion.

es6.md
```diff
@@ -189,3 +189,35 @@ interface ImportNamespaceSpecifier {
 ```
 
 A namespace import specifier, e.g., `* as foo` in `import * as foo from "mod.js"`.
+
+## ExportDeclaration
+
+```js
+interface ExportDeclaration <: Node {
```
jeffmo: I always feel like I have to sit down and think really hard about this structure because it represents so many different semantic forms. I'm always worried about screwing it up because it has so many states to consider -- many of which are even invalid in unobvious ways: 

* Having `default` be true and `source` be non-null is, currently at least, invalid
* Having `default` be true and `specifiers` be an empty list -- or a list of more than 1 -- is invalid
* Having multiple BatchSpecifiers in a specifiers list is invalid
* Having a BatchSpecifier in a list with *anything* else is invalid
* Having a BatchSpecifier with `default=true` is invalid

(there are probably more permutations...)

It would make this so much easier to reason about if we had a node-type for each of those forms (something like `ExportBatchDeclaration`, `ExportDefaultDeclaration`, `ExportNamedDeclaration`).

I don't think having zero invalid states is a goal we should shoot for in the general case, but when the number of invalid states is so high (and relatedly difficult to reason about), I think it becomes a problem.

mikesherov: @jeffmo is it possible to defer that discussion to a future PR / improvement? For the interest of documenting potential interop?

jeffmo: Aren't we trying to flesh out a structure that makes the most sense here? Seems it'd be a shame if we got this in only to decide later to come back and make a second breaking change for various places.

mikesherov: fair enough.

mikesherov: @RReverser @ariya @sebmck @impinball thoughts here on what @jeffmo is proposing?

caridy: @jeffmo I feel the same, that's why I wanted to have a more specific `declaration`, that could lead to a more strict constructor for the tree. I guess the grammar itself is the root of the evil here. I thought on splitting `ExportDeclaration` into `ExportDeclaration` and `ExportDefaultDeclaration`, but I guess that will not help that much, still we will have a lot of invalid states.

jeffmo: Still a few, but you isolate them into their own cases (and therefore reduce the number of invalid states you have to think of when dealing with any one given node):

```
interface ExportNamedDeclaration <: Node {
  declaration: Declaration | null;
  specifiers: [ ExportSpecifier ];
  source: Literal | null;
}
```

The only invalid states are having both `declaration` and `specifiers` populated and having a non-empty `specifiers` list with a non-null `source` value. Still a few invalid states, but many fewer than in the combined scenario.

```
interface ExportDefaultDeclaration <: Node {
  declaration: Declaration | Expression;
}
```

The only invalid state possible here is allowing function or class expressions (depending on who you ask :p )

```
interface ExportBatchDeclaration <: {
  source: Literal
}
```

Zero invalid states here

dherman:
:+1:

es6.md
caridy: yes, I can roll with that. I'm curious what others think about this.

sebmck: Sounds really good to me. Only possible problem I see is making code generation slightly more complicated but that's a pittance in comparison to making it easier to traverse/transform.

mikesherov: @rreverser, thoughts about what @jeffmo proposes? Would like to get confirm here so we can finalize modules!

RReverser: Sounds good to me, but better to have PR updated (another one created?) to look at entire finalized proposal.

mikesherov: Yes, agreed. @caridy, can you update PR this proposal from @jeffmo ?

caridy: yeah, will work on that today.

es6.md
```diff
@@ -189,3 +189,48 @@ interface ImportNamespaceSpecifier {
...
+interface ExportNamedDeclaration <: Node {
+    declaration: Declaration | null;
+    specifiers: [ ExportSpecifier ];
+    source: Literal | null;
+}
+```
+
+An export named declaration, e.g., `export {foo, bar};`, `export {foo} from "mod";` or `export var foo = 1;`.
+
+_Note: Having `declaration` populated with non-empty `specifiers` or non-null `source` results in an invalid state._
```
caridy: added a note about the possible invalid states.

caridy:
I think this is great progress, it looks so much nicer now, it will require some considerable changes across the board, but I think it is worth it.

es6.md
```diff
@@ -189,3 +189,48 @@ interface ImportNamespaceSpecifier {
...
+
+```js
+interface ExportDefaultDeclaration <: Node {
+    declaration: Declaration | Expression;
+}
+```
+
+An export default declaration, e.g., `export default function () {};` or `export default 1;`.
+
+## ExportBatchDeclaration
```
jeffmo: Consider this a nit: `ExportStarDeclaration` might be a slightly more clear name at this point.

jeffmo:
This is really shaping up nicely!

es6.md
mikesherov: @jeffmo whatever the spec calls it is what I'd want.

jeffmo: The spec doesn't really have a name for it except "export * /FromClause/"

jeffmo: Anyway I suppose nows the time to get +1s/-1s from everyone (@sebmck, @rreverser, @nzakas, @impinball, etc)

jeffmo: Just spoke to @ariya and a few others and they suggested `ExportAllDeclaration` is less literal and perhaps more clear than `ExportStartDeclaration` -- which is fine with me. So allow me to update my bikeshed to suggest that instead.

Looking for +1s/-1s from everyone so we can move forward with the change, or leave as-is.

ariya: +1 for `ExportAllDeclaration`.

mikesherov: :+1: for `ExportAllDeclaration`

caridy: I'm OK with `ExportAllDeclaration`, but keep in mind that we have a new feature coming up, which is exporting default from another module, where is that going to be allocated? It will be something like `export default from "mod"`, today the only way to do that is by doing `export {default} from "mod";`. I wonder if that will belong to `ExportBatchDeclaration` or `ExportDefaultDeclaration` since it is technically a batch operation with one entry to be re-exported.

jeffmo: @caridy: My immediate reaction is that it would make sense to just extend `ExportDefaultDeclaration` to include a nullable `source` property

caridy: Ok, makes sense. Let's go with `ExportAllDeclaration` instead of `ExportBatchDeclaration` then.

caridy:
I think this is ready to be merged @mikesherov, you can take care of that :)

mikesherov:
Changed considerably since last thumb from @rreverser or @sebmck . Would like explicit :+1: from either of them here. Then I can safely merge. Great work @caridy

sebmck:
Oh damn, this is clean. :+1:

RReverser:
Looks good to me. :+1:

mikesherov:
Hooray!

mikesherov:
Thanks everyone!

PR #39, opened 2015-03-03
ImportDeclaration: no longer inherit from Declaration

mikesherov:
See https://github.com/estree/estree/pull/38#discussion_r25689216

This requires no code or data structure changes from Esprima, SpiderMonkey, nor Acorn. 

Thoughts? /cc @sebmck @impinball @RReverser @ariya @dherman ?

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -151,7 +151,7 @@ interface ClassExpression <: Class, Expression {
 ## ImportDeclaration
 
 ```js
-interface ImportDeclaration <: Declaration {
+interface ImportDeclaration <: Node {
     specifiers: [ ImportSpecifier ];
     source: Literal;
 }


caridy:
:+1:

jeffmo:
+1

dherman:
:+1:

mikesherov:
Just need @sebmck or @RReverser here, and I can merge.

RReverser:
@mikesherov I probably lost the context - could you please remind what does this give to us?

mikesherov:
@rreverser, it allows the definition of `ExportDeclaration.declaration` to be: `Declaration|Expression|null` instead of having to specify all the different declarations that aren't import or export declarations. This also matches the specs definition of a declaration, in that it excludes import and export declarations.

RReverser:
> it allows the definition of `ExportDeclaration.declaration` to be: `Declaration|Expression|null`

But this PR is for `ImportDeclaration`, isn't it? How does it affect `ExportDeclaration.declaration`?

mikesherov:
Yes, this changes ImportDeclaration to not extend from `Declaration` which allows the type of `ExportDeclaration.declaration` to be `Declaration`.

If `ImportDeclaration` extends from `Declaration`, then you can't say `ExportDeclaration.declaration` allows `Declaration` because that would allow nonsense like: `export default import * from foo;`

RReverser:
Changing parent of `ImportDeclaration` doesn't change what it can import or what can `ExportDeclaration` export, so I don't see how `export default import * from foo;` example is related to this (but I might be missing something).

This might be different in semantic spec, but in AST, `Declaration`-derived nodes are those that create local bindings (declarations) - `VariableDeclaration`, `FunctionDeclaration`, and `ImportDeclaration` perfectly matches those.

mikesherov:
@rreverser, it's not about what Import can import. It's only about the list of allowed types in `ExportDeclaration.declaration`. 

If ImportDeclaration and ExportDeclaration are considered children of `Declaration` (which they aren't semantically nor does making them such in the AST offer any advantage considering declarations don't add any fields) then when we specify the values for `ExportDeclaration.declaration` wed have to say: `declaration: variableDeclaration|ClassDeclaration|FunctionDeclaration|Expression|null` instead of just `declaration: Declaration|Expression|null`.

mikesherov:
Also, you can't just place an ImportDeclaration everywhere a declaration is valid. Having `ImportDeclaration` extend `Declaration` violates Liskov substitution principle, which is important to mutating consumers of the AST, like Babel.

RReverser:
I agree that `ExportDeclaration` should not to derive from `Declaration`, as it should have real declarations in `declaration` field, but `ImportDeclaration` is different: in opposite to `ExportDeclaration` which just wraps other declarations, `ImportDeclaration` is declaration itself as it introduces ("declares") local bindings with predefined values.

RReverser:
> Also, you can't just place an ImportDeclaration everywhere a declaration is valid. Having ImportDeclaration extend Declaration violates Liskov substitution principle, which is important to mutating consumers of the AST, like Babel.

I guess I finally got what you mean. Thanks for explanations, makes perfect sense now.

mikesherov:
> And sorry for taking time, just needed to figure out reasoning behind)

No worries. That's why we're all here! I'm going to merge this now.

impinball:
Sorry for that oversight on my previous PR... Thanks. ;)

PR #47, opened 2015-03-08
Define `SpreadElement`

### LINKED ISSUE #7: Formalize SpreadElement
Estree currently doesn't define `SpreadElement`, which is used by esprima and acorn in `ArrayPattern`s and `ArrayExpression`s.

[Example](http://felix-kling.de/esprima_ast_explorer/#/ykp8P3Yy1c/1):

```js
// [a, ...b] = x;
{
  "type": "AssignmentExpression",
  "operator": "=",
  "left": {
    "type": "ArrayPattern",
    "elements": [{
      "type": "Identifier",
      "name": "a"
    }, {
      "type": "SpreadElement",
      "argument": {
        "type": "Identifier",
        "name": "b"
      },
    }],
  },
  "right": {
    "type": "Identifier",
    "name": "x"
  }
}
```

Edit:

I'm not suggesting/implying that it has to be `SpreadElement` or that it has to be used in both cases, just we have to properly define it (them) for `ArrayPattern`s and `ArrayExpression`s.

RReverser:
Fixes #7

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -31,6 +31,23 @@ extend interface VariableDeclaration {
 # Expressions
 
 ```js
+interface SpreadElement <: Node {
+    type: "SpreadElement";
+    argument: Expression;
+}
+
+extend interface ArrayExpression {
+    elements: [ Expression | SpreadElement | null ];
+}
+
+extend interface CallExpression {
+    arguments: [ Expression | SpreadElement ];
+}
+```
+
+Spread expression, e.g., `[head, ...iter, tail]`, `f(head, ...iter, ...tail)`.
+
+```js
 extend interface AssignmentExpression {
     left: Pattern | Expression;
 }
--- a/es6.md
+++ b/es6.md
@@ -47,6 +47,8 @@ extend interface CallExpression {
 
 Spread expression, e.g., `[head, ...iter, tail]`, `f(head, ...iter, ...tail)`.
 
+**FIXME:** This describes the Esprima and Acorn behaviors, which is not currently aligned with the SpiderMonkey behavior.
+
 ```js
 extend interface AssignmentExpression {
     left: Pattern | Expression;


RReverser:
@mikesherov @ariya @dherman Can you please :+1: ?

mikesherov:
Do we have to do anything special for use of `speadElement` as part of a destructuring assignment?

mikesherov:
Also, a general trend I've noticed is the lack of prose describing what these nodes represent. We've merged a few PRs without prose, but I'd like us to get away from that.

Also, do we have to specifically call out that the redefinition of CallExpression and ArrayExpression here? Meaning, maybe an informative note mentioning the difference and that this definition overrides the ES5 definition?

RReverser:
In destructuring assignment it's `RestElement`, and for cases like `someFunc(...x)` and `var ar = [1, ...otherArr, 2]` it's `SpreadElement`.

RReverser:
> Also, do we have to specifically call out that the redefinition of CallExpression and ArrayExpression here? Meaning, maybe an informative note mentioning the difference

It's not redefinition - we already use `extend interface ...` in this DSL in every place where we want to add / redefine particular properties with new semantics.

RReverser:
Just another note - it's not something new, this is how things already work in Acorn and Esprima.

RReverser:
As for SpiderMonkey, it seems to use something like following instead:

```js
interface SpreadExpression <: Node {
    expression: Expression;
}
```

(need @dherman to confirm)

es6.md
```diff
@@ -31,6 +31,18 @@ extend interface VariableDeclaration {
 # Expressions
 
 ```js
+interface SpreadElement <: Node {
+    argument: Expression;
```
Constellation: It also requires `type: "SpreadElement";`, correct?

RReverser: Yes, thanks, @Constellation!

mikesherov:
I'll defer to @ariya here.

RReverser:
Would be nice to know opinions of other participants, too.

dherman:
That's right re: SpiderMonkey:
```
js> Reflect.parse("f(1,...a,2)", { loc: false }).body[0].expression.arguments[1]
({loc:null, type:"SpreadExpression", expression:{loc:null, type:"Identifier", name:"a"}})
```
I'll try to submit a patch to fix it soon.

dherman:
I agree that `extend interface` is enough to call out backwards-compatible extensions.

I agree with @mikesherov that there should be prose for each element. :+1: from me for this PR otherwise.

RReverser:
Added prose and warning about SM behavior.

RReverser:
@ariya Please give :+1: so I could merge this.

ariya:
:+1:

PR #50, opened 2015-03-08
Add type field to RestElement and Assignmentpattern

Constellation:
This patch adds type field to RestElement and AssignmentPattern since
they can be allocable nodes.

Could you take a look? /cc @RReverser


RReverser:
Right, this is definitely smth missed, so I'm merging it right away.

PR #56, opened 2015-03-10
Restructure experimental docs

sebmck:
This should be a non-controversial change. Just need a couple :+1: and I'll merge. I plan on submitting PRs for some upcoming specs (declarators, class property initializers) so this is more scalable.


michaelficarra:
LGTM

sebmck:
Merging this since it shouldn't be very controversial as it's a non-additive change.

PR #57, opened 2015-03-12
Add SuperExpression

### LINKED ISSUE #54: Do we need SuperExpression?
For `this`, there's `ThisExpression` in the AST.

Do we need a `SuperExpression` to represent calls to `super`?

RReverser:
Fixes #54

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -31,12 +31,14 @@ extend interface VariableDeclaration {
 # Expressions
 
 ```js
-interface SuperExpression <: Expression {
-    type: "SuperExpression";
+interface Super <: Expression {
+    type: "Super";
 }
+```
 
-A `super` expression.
+A `super` pseudo-expression.
 
+```js
 interface SpreadElement <: Node {
     type: "SpreadElement";
     argument: Expression;
--- a/es6.md
+++ b/es6.md
@@ -31,6 +31,12 @@ extend interface VariableDeclaration {
 # Expressions
 
 ```js
+interface SuperExpression <: Expression {
+    type: "SuperExpression";
+}
+
+A `super` expression.
+
 interface SpreadElement <: Node {
     type: "SpreadElement";
     argument: Expression;


RReverser:
Need to get confirmation from Esprima team to get this merged, as they're the only who didn't reply in original issue. /cc @mikesherov @ariya @caridy

mikesherov:
I'll defer to @ariya here. Should be simple enough to add....

michaelficarra:
This is a backward incompatible change for a parse that is already agreed upon by all major implementations. How can this be considered given the goal of this project?

mikesherov:
Because we can consider things without agreeing to actually do them, @michaelficarra . I'm not sure why you keep assuming what we are allowed to do or not do. All we have said from the beginning is that we won't release BC breaking changes for ES5, and that we're here to primarily reach agreement on ES6.

Our readme says:

```
Meanwhile JavaScript is evolving, notably with the upcoming release of ES6. This site will serve as a community standard for people involved in building and using these tools to help evolve this format to keep up with the evolution of the JavaScript language.
```

Also, we all can release new major versions of our parsers if we agree to the change, a.k.a. what we did with RestElement.

michaelficarra:
Okay, I didn't know there was a difference in how we were treating ES5/ES6 features. That makes a little more sense. I thought we were optimising 100% for not breaking backward compatibility.

mikesherov:
@michaelficarra sorry for the aggressive tone. I should have made that more clear. I'm submitting a PR tonight that adds guiding principles to the readme, because there does seem to be a bit of confusion around this topic in general. Thanks for sticking with me!

nzakas:
So... good to merge? :)

mikesherov:
Waiting for feedback from @dherman and @ariya .

nzakas:
@dherman @ariya thoughts?

RReverser:
I asked @dherman to look at this, might be too busy yet.

ariya:
LGTM. From Esprima perspective, it will be in its next major version (we already shipped `class` without a special treatment for `super`).

mikesherov:
We want this, but we want the name not to be `SuperExpression` but rather `Super` considering it's not an expression. The parity with `thisExpression` is a red herring.

sebmck:
@RReverser Looks like once this is renamed to `Super` it's good to merge.

gibson042:
@sebmck Does that mean you backed off on `SuperProperty` and `SuperCall`?

sebmck:
@gibson042 I don't really care, it was just a suggestion. Not advocating for or against it.

RReverser:
Dunno... In general `Super` sounds fine (while somewhat doesn't fit into most of ESTree types), and looks like the only difference with any other expression is contextual limits which might very well change in future so that for example `Object.keys(super)` and other things will be allowed. Anyway, I'll update it now and will be waiting for :+1: to make sure everyone is on the same page.

nzakas:
:+1: 

@Constellation

mikesherov:
Seems we must summon someone from SM here to get a :+1: ?

mikesherov:
ah, @efaust has said he'll follow our lead here. We should merge this.

PR #58, opened 2015-03-12
Allow nesting MemberExpression in Patterns

### LINKED ISSUE #49: MemberExpression is not allowed inside Patterns
With current spec, following code examples cannot be represented correctly:

- `[...obj.prop] = arr`
- `[this.x] = arr`
- `{a: this.x} = other`
- *...and others*

Thinking what is the best way to fix this.

Is it adding `Pattern` as parent to `MemberExpression`? Or will it be confusing in contexts like `VariableDeclarator.id` and `Function.params[]`?

RReverser:
Fixes #49

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -524,10 +524,10 @@ A `new` expression.
 ## MemberExpression
 
 ```js
-interface MemberExpression <: Expression {
+interface MemberExpression <: Expression, Pattern {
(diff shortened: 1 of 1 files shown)


RReverser:
Since this is something that already works and requires zero code change, I'll just merge it.

PR #59, opened 2015-03-12
Add sourcetype to Program node in ES6 (closes #46)

### LINKED ISSUE #46: Add flag on Program node to indicate module code?
It would be nice to be able to tell if a given Program represents module code or not. The Shift AST actually has two different nodes `Script` and `Module` to represent the source type. It's probably too big of a change to go for a separate node in ESTree, however, what about a property on `Program`? Maybe `sourceType` as "script" or "module"?

nzakas:
Adding `sourceType` to `Program` node.

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -1,5 +1,15 @@
 This document specifies the extensions to the core ESTree AST types to support the ES6 grammar.
 
+# Programs
+
+```js
+extend interface Program {
+    sourceType: "script" | "module";
+}
+```
+
+Parsers must specify `sourceType` as `"module"` if the source has been parsed as an ES6 module. Otherwise, `sourceType` must be `"script"`.
+
 # Functions
 
 ```js


caridy:
:+1:

getify:
+1

es6.md
```diff
@@ -1,5 +1,15 @@
 This document specifies the extensions to the core ESTree AST types to support the ES6 grammar.
 
+# Programs
+
+```js
+extend interface Program {
+    sourceType: "script" | "module"
```
RReverser: Grammar nazi mode: please put a semicolon here :)

Other than that, :+1:

nzakas: Heh, my pleasure.

nzakas:
Updated. Too bad there's not a linter. ;-)

nzakas:
Ready for merge?

RReverser:
Probably needs feedback from @ariya and @dherman.

mikesherov:
:+1:

michaelficarra:
:+1:

jeffmo:
Big :+1:

jeffmo:
In that past I'd also considered the thought of just having a `Program` (for backward compat) and a `Module` toplevel container. I suppose this works just as well, it just means the limitations on the contained statements in the body of `Program` has to be based on the property of the toplevel object (rather than its type).

jeffmo:
FWIW it's likely that modules will continue to have a unique set of valid statements they can contain -- for example there's discussion about `await import` brewing (and toplevel `await` in general in a module) since modules are loaded asynchronously already.

RReverser:
Well, yeah, but obviously it's not the only place where allowed set is different - for example, any `Function` node with `generator: true` can contain `YieldExpression`s, while any other can't.

This one doesn't seem to harm any processing / checks, so `sourceType` looks reasonable to me.

jeffmo:
That's true. To be clear: I'm not suggesting you can necessarily do something with one over the other (nor am I suggesting that I'm in any way opposed to this PR) -- just throwing the differences out there in case it flags something for other consumers.

dherman:
:+1: -- I had the same initial instinct as @jeffmo but I think this is an elegant way to pretend I meant to do it this way all along. ;-P

mikesherov:
Thanks everyone!

Issue #41, opened 2015-03-04
Concrete Syntax in tree


getify:
I'd like to open up the discussion of what our various options are for preserving so called "Concrete Syntax" elements in the parse tree format.

## What are Concrete Syntax elements?

A simplified definition of "concrete syntax" is any stuff that could appear in a source code program which, when it's parsed, would be otherwise discarded information that an AST (as it stands now) does not represent.

That specifically means that things which are reliably inferable from an AST's nodes are *not* concrete syntax. For example:

```js
var a = 2,
    b = (a + 2) * 3;
```

Here, the `( )` around `a + 2` is not represented in the AST, because the structure of the tree, combined with operator precedence rules, absolutely implies that it must exist, and moreover `a + 2 * 3` would have been a different tree structure. Thus, the `( )` in the original program can be reconstructed from the AST reliably, so it need not be stored separately. **It is not concrete syntax**.

However:

```js
var a = 2,
    b = a + (2 * 3);
```

This program includes a pair of `( )` that would already be implied by the tree structure and operator precedence, and thus would also not be stored.

But, critically, they also would not be re-generated when the tree was reconstituted. Why? Because it's impossible to know from the tree alone if the `( )` was really there, or just implied. And code generation takes the conservative path and doesn't make up `( )` where it's not sure they were there, so it leaves them out.

Herein we see that the original `( )` are not abstract syntax, but concrete, and to preserve them is going to require some other solution.

Examples of concrete syntax:

* extraneous whitespace (the non-significant kind)
* comments
* extraneous `( )` (used primarily for readability more than functionality)
* ... other?

### The Rabbit Hole

Just how deep does this rabbit hole go? In the following snippet, every single `/*x*/` comment represents a location where whitespace and/or comments can optionally appear as concrete syntax, in addition of course to the fact that some of those locations require significant whitespace, such as `/*4*/`:

```js
/*1*/class /*2*/ Foo /*3*/ extends /*4*/ Bar /*5*/ {
   /*6*/constructor /*7*/ (/*8*/x /*9*/ = /*10*/ "hello" /*11*/) /*12*/ {
      // ..
   }

   /*13*/ static /*14*/ bam /*15*/ ( .. ) {
      // ..
   }
}

var /*16*/{ /*17*/ x /*18*/: /*19*/ { /*20*/ y /*21*/: /*22*/ z /*23*/ = /*24*/ 2 /*25*/ } } =
[...]

Issue #14, opened 2015-02-13
.raw property of literal nodes, or something similar


dherman:
Discussed some in #6. The `raw` property of literal nodes is an Esprima extension.

I have a specific use case for the `raw` property, or at least for more information than is provided by `value`: in asm.js, the distinction between an int literal and float literal is significant and distinguished by the type system. Writing an asm.js validator in JS requires the ability to distinguish between e.g. `17.0` and `17`, but the `value` property does not distinguish these.

OTOH, @michaelficarra points out that AST nodes might be produced by tools that don't want to have to specify the raw source. A few options I can see:

* just leave `raw` unspecified and treat asm.js as a special case that is adequately served by Esprima's extended behavior
* add a spec for additional node data to be produced by parsers but optional for other tools, and include `raw` in that
* specify just the coarse-grained "type" of the lexeme, like `type: "int" | "float" | "string" | "boolean" | "null" | "RegExp"`
* specify finer-grained lexical class information, something like `type: "hex" | "octal" | "decimal" | "float" | "boolean" | "null" | "RegExp"`

I am hesitant to generalize based on the one use case of asm.js. Without knowing of other use cases, I think the most conservative step is just to document the existing practice of `raw` in an optional spec.

Thoughts?

RReverser:
Acorn also produces `raw` property, but I'm not sure if it's something that should be standardized as a "must" for all the parsers as raw representation can be easily retrieved using original code string + Literal's `range` property.

ariya:
@RReverser That requires the node location to be specified. A code generator that takes an AST may not need the location at all, but it can benefit from `raw`.

RReverser:
@ariya Makes sense, thanks for a good example.

michaelficarra:
Remember, `SourceLocation` (which all `Node`s may have)

    interface SourceLocation {
        source: string | null;
        start: Position;
        end: Position;
    }

has an optional `source`. Unfortunately, it would require you to have a start/end position listed as well. I think specifying esprima's `raw` member is fine.

Also notable, @RReverser: `range` is not part of `SourceLocation`. It is yet another esprima extension.

RReverser:
@michaelficarra Yes, I already agreed that possible benefits of `raw` are clear :)

mikesherov:
This delves into the realm of a CST, which has lots of interested parties... Perhaps we can defer this discussion until we have a comprehensive discussion of what a concrete syntax tree would look like?

getify:
[...]

PR #63, opened 2015-03-14
Add `kind`

### LINKED ISSUE #61: `kind` field for Literal
It is very useful for basic type checking, and IMO is a bit less hackish (and easier) than `typeof node.value`. Here's my idea: 

```js
interface Literal <: Node, Expression {
  type: "Literal";
  value: string | boolean | null | number | RegExp;
  // addition, the type
  kind: "string" |
    "boolean" |
    "null" |
    "undefined" |
    "number" |
    "regexp";
}
```

impinball:
Fixes #61.

`kind === "undefined"` was ignored from the original proposal since `undefined` is represented by an Identifier, not a Literal.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -593,10 +593,11 @@ An identifier. Note that an identifier may be an expression or a destructuring p
 interface Literal <: Node, Expression {
     type: "Literal";
     value: string | boolean | null | number | RegExp;
+    kind: "string" | "boolean" | "null" | "number" | "regexp";
 }
 ```
 
-A literal token. Note that a literal can be an expression.
+A literal token. Note that a literal can be an expression. `kind` represents the type of `value`. Examples include `kind === "string"` if `value === "str"` or `kind === "number"` if `value === 2`.
 
 ### RegexLiteral
 
--- a/spec.md
+++ b/spec.md
@@ -593,7 +593,7 @@ An identifier. Note that an identifier may be an expression or a destructuring p
 interface Literal <: Node, Expression {
     type: "Literal";
     value: string | boolean | null | number | RegExp;
-    kind: "string" | "boolean" | "null" | "number" | "regexp";
+    kind: "string" | "boolean" | "null" | "number" | "regex";
 }
 ```
 


spec.md
```diff
@@ -593,10 +593,11 @@ An identifier. Note that an identifier may be an expression or a destructuring p
 interface Literal <: Node, Expression {
     type: "Literal";
     value: string | boolean | null | number | RegExp;
+    kind: "string" | "boolean" | "null" | "number" | "regexp";
```
gibson042: 27d0e173d7b220044c5fa05da1e9c3cfcf96a41b already established a precedent of referring to regular expressions with `regex`, which we should maintain for internal consistency.

gibson042:
I hate saying this, but I feel like the "[`kind`](https://github.com/estree/estree/issues/61)", "[`raw`](https://github.com/estree/estree/issues/14)", [concrete syntax](https://github.com/estree/estree/issues/41), and [(JSON) portability](https://github.com/estree/estree/issues/new) issues are all inextricably entangled by our reasonable goal of avoiding redundancy for simplifing mutations and minimizing memory footprint. "`value`" adds—well, value—but is not always portable, is incomplete in the absence of source input elements, and is highly redundant (but I suppose I'm showing my bias for https://github.com/estree/estree/issues/14#issuecomment-77826124 ).

This should be adopted now _only_ if we're willing to impose the corresponding memory and complexity (i.e., cross-property synchronization) demands, and otherwise held until those other threads catch up.

sebmck:
I've been thinking about this more and more and I'm starting to be a heavy proponent against this. It's not that hard to check if a literal value is of a certain type (`typeof node.value === "string"`, it's even simpler with a utility library) and having this type of metadata that can easily be out of sync is dangerous.

RReverser:
> and having this type of metadata that can easily be out of sync is dangerous

I agree with this concern. The only thing that couldn't be checked easily was `regex`, but it's check is now covered by presence of extra property.

gibson042:
Note that `regex` and `value` are theirselves "metadata that can easily be out of sync".

RReverser:
Yes, but those are for a good reason here - it's not just about conveniency, but rather about impossibility to represent certain regexes in current&older engines.

impinball:
Although it might not make sense to all of us used to checking `typeof`, but this would also play nicely with #64, which adds a `raw` field to it as well. The latter is especially nice WRT minifiers and ASM.js tools (float vs int). It's also especially useful for consumers that aren't using JavaScript/friends or a language that can easily interface with it, such as C++. A good example of this would be Python, in which because of language differences, it's awkward to bridge with JavaScript. Consider the problems JeanSebTr/node-python has, and that it's the best out of very few in existence. JSON would be easier to deal with here.

impinball:
@sebmck I'm well used to checking `typeof` myself. I'm just looking at the fact not all tools use JavaScript.

@gibson042 Fixed the `regex`/`regexp` inconsistency.

gibson042:
> Although it might not make sense to all of us used to checking `typeof`, but this would also play nicely with #64, which adds a `raw` field to it as well.

I strongly agree with this. `kind` + `raw` vastly improve portability, and are valuable even within Javascript.

RReverser:
> I'm well used to checking typeof myself. I'm just looking at the fact not all tools use JavaScript

There is a lot of places in AST where dynamic tools in language are meant to be available as we have type unions, optional extensions and so. So language being able to determine string from number is a must.

> `kind` + `raw` vastly improve portability

...while breaking interoperability at the same time. `kind` wasn't present in pre-ES6 spec so some tools might continue to simply change `value` of `Literal` to different type (while not updating unknown `kind` field), while new tools will already rely on the `kind`, despite that it's not synchronized.

impinball:
@RReverser 

> > kind + raw vastly improve portability
>
> ...while breaking interoperability at the same time. kind wasn't present in pre-ES6 spec so some tools might continue to simply change value of Literal to different type (while not updating unknown kind field), while new tools will already rely on the kind, despite that it's not synchronized.

:+1: 

On the other hand, is there any sort of deprecation process? I think we could get away with adding both `kind` and `raw`, while deprecating `value`. IMHO it would be easier and more consistent to handle, and getting the `value` equivalent is still a relatively simple function. It would also be easier to deal with for JSON consumers, and would completely rectify #64.

```js
function getValue(literal) {
  if (literal.type !== 'Literal') throw TypeError('node is not a Literal');
  switch (literal.kind) {
  case 'string': return literal.raw.slice(1, -1);
  case 'number': return +literal.raw;
  case 'boolean': return literal.raw === 'true';
  case 'null': return null;
  case 'regex':
    return RegExp.apply(null,
      [].slice.call(/^\/(.+)\/([a-z]*)/.exec(literal.raw), 1));
  }
}

function buildNodeFromLiteral(value) {
  switch (typeof value) {
  case 'object':
    if (value !== null && [].toString.call(value) !== '[object RegExp]') {
      break;
    }
  case 'number':
  case 'boolean': return makeLiteral(value + '');
  case 'string': return makeLiteral('"' + value.replace(/"/g, '\\"') + '"');
  }
  throw TypeError('value cannot be converted to a literal');
}
```

sebmck:
@impinball 

> On the other hand, is there any sort of deprecation process?

There's no deprecation process because nothing is going to be deprecated.

> IMHO it would be easier and more consistent to handle, and getting the value equivalent is still a relatively simple function

Getting any of this metadata is relatively easy, I'm not sure why you're advocating changing this just for the sake of it. There are other more apt solutions for the JSONability.

sebmck:
Here's an example. I want to check if a string is of a certain value without any helper methods. Here's what I now have to do:

```javascript
myString.kind === "string" && (myString.raw === '"my string"' || myString.raw === "'my string'");
```

Unnecessarily convoluted.

impinball:
@sebmck You have a good point.

@all This okay to close now, since it's starting to sound better on paper than in practice?

michaelficarra:
@sebmck not only that, you have to interpret escape sequences.

@impinball yes, I think the other open issues have more focus.

PR #67, opened 2015-03-19
Add philosophy to README

sebmck:
This is my first stab at an ESTree "philosophy". So far it's a collection of points taken from previous issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,3 +16,12 @@ We've started the process of bringing together various communities using this fo
 * Michael Ficarra ([@michaelficarra](https://github.com/michaelficarra))
 * Sebastian McKenzie ([Babel](https://github.com/babel/babel))
 * Kyle Simpson ([@getify](https://github.com/getify))
+
+# Philosophy
+
+Suggested additions and modifications must follow these guidelines:
+
+1. **Backwards compatible:** Non-additive modifications to existing constructs will not be considered unless immense support is in favor of such changes. ([eg. #65](https://github.com/estree/estree/issues/65))
+2. **Contextless:** Nodes should not retain any information about their parent. ie. a `FunctionExpression` should not be aware of if it's a concise method. (eg. [#5](https://github.com/estree/estree/issues/5))
+3. **Unique:** Information should not be duplicated. ie. a `kind` property should not be present on `Literal` if the type can be discerned from the `value`. (eg. [#61](https://github.com/estree/estree/issues/61))
+4. **Extensible:** New nodes should be specced to easily allow future spec additions. This means expanding the coverage of node types. ie. `MetaProperty` over `NewTarget` to cover future meta properties. (eg. [#32](https://github.com/estree/estree/pull/32))


gibson042:
Sorry, I didn't see this before posting my diatribe to #32. I've copied the relevant portions below, and will see if I can recast them into components of this PR.

TL;DR:
* Restrict AST structure _only_ by typing property values in type definitions
  * …which is blind to recursive "contains/does-not-contain" syntax violations
* Prefer updating existing types over creating new subtypes
* Prefer creating new subtypes over copying and tweaking existing types

> It's up to the spec to help define what is a valid program. If it weren't, then we could just make everything extend from Node and not worry about typing at all. Our type system is there to enforce program validity. To suggest otherwise is against the spirit of a spec to begin with. ESTree takes a more lax approach than Shift on principle…

ES6 has two major classes of syntax errors: those dependent upon token values (e.g., exceeding the upper bound for valid `\u{ HexDigits }` escape sequences) and those dependent upon code structure (e.g., `break LabelIdentifier` where LabelIdentifier is not in the containing label set). Even errors in the first category are not _actually_ independent of structure (e.g., they can apply only in strict mode, which can be opted into by a "use strict" directive higher in the scope), but I'll ignore that for the moment.

ESTree has shown no intention to guarantee absence of the first class, and cannot guarantee absence of the second class in a general sense. It can, however, rule out patently invalid constructs (e.g., defining the label of BreakStatement to be an Identifier precludes trees in which label is a more general Expression—or at least provides a mechanism to identify such trees as invalid), which I think is what you mean by "Our type system is there to enforce program validity". Such restrictions seem limited to type definitions, which relate only to direct containment, but if they are not now so limited then I would propose that as a fundamental **principle**. This in fact relates to Liskov substitution—where a container property allows a type, it must necessarily allow all subtypes. Note also that it only goes one level down, so we're incapable of e.g. disallowing continue statements outside of iteration statements unless we start splitting Statement, resulting in a proliferation of types.

Speaking of which, the topics of subtypes and proliferation are also crying out for principles. There's a balancing act with both, a desire to be sufficiently descriptive for consumers without splitting hairs _so_ finely that it is burdensome to deal with a multitude of node types that are indistinguishable for a particular purpose. I believe addressing that latter concern is the value of subtypes: to introduce supplemental variation that can be safely ignored unless it is relevant to the operation at hand. From this perspective, `Statement` subtypes are decent and `Expression` subtypes are a little worse, but moving in the wrong direction. So, on to the proposed **principles**: 1) since subtypes are defined by their inheritance (which ultimately controls only the Node (interface, property) pairs in which they are valid) and property list, I propose that updating an existing type (through some combination of property value alteration and property addition) be preferred over new subtype creation unless doing so would allow a syntactically invalid or meaningless construct to be expressed as a valid tree, excepting those errors that could be masked by replacing direct containment with indirect (e.g., by wrapping an Expression in a SequenceExpression); and 2) to avoid duplication in both this spec and in consumers of it, subtyping should be preferred over type copying when one type's properties are a (possibly non-strict) subset of another type.

mikesherov:
@sebmck can you mention that ES5 is frozen 100%. The language about BC seems to imply that ES5 can change under extreme cases.

mikesherov:
:+1:

mikesherov:
@gibson042 I think it's fine for us to discuss your points. Perhaps we can get the basics here merged in before we get onto thornier, more technical philosophies like you are proposing?

RReverser:
Yup, looks good to me, I had only one more suggestion - maybe worth keeping in relevant CONTRIBUTING.md so that Github shows it in bar over each new issue?

README.md
```diff
@@ -16,3 +16,12 @@ We've started the process of bringing together various communities using this fo
...
 * Kyle Simpson ([@getify](https://github.com/getify))
+
+# Philosophy
+
+Suggested additions and modifications must follow these guidelines:
+
+1. **Backwards compatible:** Non-additive modifications to existing constructs will not be considered unless immense support is in favor of such changes. ([eg. #65](https://github.com/estree/estree/issues/65))
+2. **Contextless:** Nodes should not retain any information about their parent. ie. a `FunctionExpression` should not be aware of if it's a concise method. (eg. [#5](https://github.com/estree/estree/issues/5))
+3. **Unique:** Information should not be duplicated. ie. a `kind` property should not be present on `Literal` if the type can be discerned from the `value`. (eg. [#61](https://github.com/estree/estree/issues/61))
+4. **Extensible:** New nodes should be specced to easily allow future spec additions. This means expanding the coverage of node types. ie. `MetaProperty` over `NewTarget` to cover future meta properties. (eg. [#32](https://github.com/estree/estree/pull/32))
```
gibson042: Suggested additions:
* **Liberal**: Nodes impose no restrictions on their children beyond type and no restrictions at all on deeper descendants. Many conforming ESTree objects represent invalid ECMAScript code, and consumers for which this is problematic must handle context-sensitive restrictions (strict mode, required/prohibited ancestor nodes, name uniqueness/validity, etc.) theirselves.
* **Portable**: Structures should be losslessly serializable into a variety of formats (but most notably, JSON), and therefore should be acyclic and utilize only object, array, string, number, boolean, and null data types.
* **Concise**: Updating existing types is preferred over introducing new types, and introducing new subtypes is preferred over introducing slightly-modified copies of existing types.

gibson042:
> @gibson042 I think it's fine for us to discuss your points. Perhaps we can get the basics here merged in before we get onto thornier, more technical philosophies like you are proposing?

I guess that depends on whether or not they are contentious, but I have no problem with merging this as-is.

RReverser:
This one left right now. Can we merge what we already have and work on future clarifications on separate PRs?

mikesherov:
Yes. Merge at will!

PR #70, opened 2015-03-19
fix Super

michaelficarra:


gibson042:
:+1: for use of CallExpression and MemberExpression. I'm on the fence about `type: "Super"`, but removing its inheritance from Expression is still an improvement.

es6.md
```diff
@@ -31,9 +31,17 @@ extend interface VariableDeclaration {
 # Expressions
 
 ```js
-interface Super <: Expression {
+interface Super {
```
gibson042: `<: Node`

RReverser:
Agree, should inherit from `Node` at least. Other than that, looks good.

michaelficarra:
Updated. Thanks for the review.

ikarienator:
LGTM :+1: Thanks!

RReverser:
LGTM

mikesherov:
LGTM. SM has said they'll follow us here, so this should be good to merge.

RReverser:
Ok, someone has to do it then :)

PR #72, opened 2015-03-29
ImportDeclaration specifiers

leebyron:
ImportDeclaration specifiers should be able to include ImportDefaultSpecifier or ImportNamespaceSpecifier


sebmck:
Thanks @leebyron!

PR #75, opened 2015-04-07
Module specifiers inherit from Node

### LINKED ISSUE #74: Module specifiers do not inherit from Node
`ExportSpecifier`, `ImportSpecifier`, `ImportDefaultSpecifier`, and `ImportNamespaceSpecifier` do not inherit directly or indirectly from `Node`. Is this intentional? I looked through #11 and #35 but couldn't find any discussion to indicate that that would be the case.

I investigated and found that jquery/esprima#1149 will attach `loc` to all four node types when requested. The ast-types implementation does not have modules as they are currently specified, but it does have specifiers inherit from [an abstract `Specifier` type](https://github.com/benjamn/ast-types/blob/master/def/es6.js#L162), which itself inherits from `Node`.

btmills:
Fixes #74

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -272,7 +272,7 @@ An import declaration, e.g., `import foo from "mod";`.
 ## ImportSpecifier
 
 ```js
-interface ImportSpecifier {
+interface ImportSpecifier <: Node {
     type: "ImportSpecifier";
     imported: Identifier;
     local: Identifier;
@@ -284,7 +284,7 @@ An imported variable binding, e.g., `{foo}` in `import {foo} from "mod"` or `{fo
 ## ImportDefaultSpecifier
 
 ```js
-interface ImportDefaultSpecifier {
+interface ImportDefaultSpecifier <: Node {
     type: "ImportDefaultSpecifier";
     local: Identifier;
 }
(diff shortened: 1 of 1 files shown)


michaelficarra:
LGTM.

RReverser:
Probably worth adding abstract base like:
```
interface Specifier <: Node {
  local: Identifier;
}
```
and inheriting from it?

impinball:
Probably a good idea.

btmills:
Updated to inherit from a new `Specifier` type.

RReverser:
LGTM!

btmills:
Is there any likelihood that in the future there will be other types of specifiers not related to modules? If that's a possibility, the new base could be named `ModuleSpecifier` instead.

RReverser:
Dunno. Theoretically, yes, but this is just an abstract interface so we can safely rename it in future without breaking most tools, but you can rename it now to `ModuleSpecifier` just to be 100% sure.

btmills:
Renamed.

RReverser:
Cool! Any other concerns before merging this, anyone?

nzakas:
:+1:

RReverser:
Merging as a non-breaking.

PR #76, opened 2015-04-09
[RFC] Suggested changes to RegExpLiteral

benjamn:
The built-in constructor is called `RegExp`, so it seems weirdly inconsistent to call this type `RegexLiteral` instead of `RegExpLiteral`.

I also think we should avoid using a nested, untyped object to represent the pattern and flags of the regular expression, because that sets an unnecessary precedent. Let's either promote that object to be a typed AST node (recall that `Property` nodes were once untyped objects, but the Esprima community found it useful for them to have an actual type), or just flatten the object into multiple fields of the `RegExpLiteral` node (I prefer this, as you can see from this PR).

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -596,10 +596,10 @@ interface Literal <: Node, Expression {
 
 A literal token. Note that a literal can be an expression.
 
-### RegexLiteral
+### RegExpLiteral
 
 ```js
-interface RegexLiteral <: Literal {
+interface RegExpLiteral <: Literal {
   regex: {
     pattern: string;
     flags: string;


RReverser:
Nested object is nice because you can do `if (object.regex)` to determine that this is `RegexLiteral`. With `if (object.pattern !== undefined)` or `if (object.flags !== undefined)` that's kinda less obvious.

I don't mind using `RegExpLiteral` - this is anyway abstract interface that is mostly useful within the spec than for tools, so it's totally safe to rename it if you wish.

As for typing of this nested object - I don't see the benefits here given that `Property` did represent separate syntactic structure and thus had to be AST node (which I believe it initially didn't become due to simple oversight), while `regex` property here doesn't represent anything new - it just provides additional meta info about the same node.

Anyway, interesting to see what others think about it. Personally I'm open to any change here, just pointing out concerns.

sebmck:
This would be a breaking change so this is out of scope for ESTree.

benjamn:
I see your point about checking for the `.regex` property being easier than checking for multiple properties.

I could also imagine testing to see if `.type === "RegExpLiteral"`, but it seems strange for some `Literal` subtypes to have their own types while others just have `.type === "Literal"`.

Maybe we should do something like we do with [`AssignmentProperty`](https://github.com/estree/estree/blob/master/es6.md#objectpattern)?
```js
interface AssignmentProperty <: Property {
    type: "Property";
    ...
}

interface RegExpLiteral <: Literal {
    type: "Literal";
    ...
}
```

benjamn:
@sebmck where should this discussion be happening, then?

sebmck:
@benjamn Nowhere really. ESTree is committed to backwards compatibility so the only way this change could happen is in a new AST spec.

RReverser:
@benjamn Dunno, in `AssignmentProperty` I just put that comment to ensure that no-one is confused (while it should be anyway clear that `type` as any other property is already inherited). If you wish, we could put it.

RReverser:
@sebmck As far as I remember from some other discussions, we treated `spec.md` as fixed and `es6.md` as subject-to-change for now. Didn't we?

RReverser:
Anyway, rename to `RegExpLiteral` seems totally fine to me as it doesn't break things, but as for `regex` literal we should either discuss it or leave as-is. As for the explicit `type` inheritance, surely you can put it if you wish.

benjamn:
Overriding the `.type` to be `"Literal"` is actually important for backwards compatibility, I think. The latest version of esprima-fb, for example, produces
```js
{ type: 'Literal',
  value: /./,
  raw: '/./',
  regex: { pattern: '.', flags: '' } } }
```
If we override the type, then it doesn't matter what we call the abstract type. I would still prefer to call it `RegExpLiteral`, but it's moot since you'll never see that type in a real AST (only `"Literal"`).

I'll update this PR with that in mind.

sebmck:
@RReverser This **is** in `spec.md`.

benjamn:
@sebmck do you see my point about needing to override the `.type` to be `"Literal"`, if we're trying to capture existing parser behavior?

benjamn:
No, wait, I'm confused. In this specification, you have to explicitly specify a new type if you want it to be different from what was inherited (which is `"Literal"` in this case already).

sebmck:
The internal name is just informative so there's no issue with changing it. I thought you were proposing an API or structural change.

benjamn:
@sebmck Oh, I definitely was proposing a structural change (flattening `.regex`), but I totally see your points about that, so I removed it from the PR.

mikesherov:
:+1: if this is simply informative and requires no code changes.

mathiasbynens:
This doesn’t affect the actual AST, so sure, why not.

sebmck:
Merging this as it's not controversial. If someone feels strongly about it then feel free to speak up.

PR #79, opened 2015-04-19
Add abstract ModuleDeclaration type

### LINKED ISSUE #78: Module declarations not allowed in Program.body
After the discussions in #38 and #39, module declarations inherit directly from `Node` instead of `Declaration`. Because `Program.body` is defined as `[ Statement ]`, this has the effect of disallowing module declarations in `Program.body`.

What is the best solution? A couple possibilities:

1. Define `Program.body` to be `[ Statement | ImportDeclaration | ExportNamedDeclaration | ExportDefaultDeclaration | ExportAllDeclaration ]`, and add a note saying module declarations are only valid when `sourceType` is `"module"`.
1. Create an abstract `ModuleDeclaration` type from which module declarations inherit, and define `Program.body` to be `[ Statement | ModuleDeclaration ]`, adding a note saying `ModuleDeclaration` is only valid when `sourceType` is `"module"`.
1. Something else entirely?

I'm happy to put together a PR after a decision is made.

btmills:
Previously, module `import` and `export` declarations inherited directly from `Node`, meaning they were not allowed within `Program.body`. This change adds an abstract `ModuleDeclaration` type from which all module declarations inherit and extends `Program.body` to allow `ModuleDeclaration`s.

Fixes #78

### DIFF
--- a/es6.md
+++ b/es6.md
@@ -5,6 +5,7 @@ This document specifies the extensions to the core ESTree AST types to support t
 ```js
 extend interface Program {
     sourceType: "script" | "module";
+    body: [ Statement | ModuleDeclaration ];
 }
 ```
 
@@ -257,6 +258,14 @@ interface MetaProperty <: Expression {
 
 # Modules
 
+## ModuleDeclaration
+
+```js
+interface ModuleDeclaration <: Node { }
+```
+
+A module `import` or `export` declaration.
+
 ## ModuleSpecifier
 
 ```js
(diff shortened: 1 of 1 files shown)


caridy:
LGTM

RReverser:
:+1:

es6.md
```diff
@@ -5,6 +5,7 @@ This document specifies the extensions to the core ESTree AST types to support t
 ```js
 extend interface Program {
     sourceType: "script" | "module";
+    body: [ Statement | ModuleDeclaration ];
```
michaelficarra: We could be more restrictive with this, allowing module declarations only when sourceType is "module".

RReverser: That could be another PR that would split `Program` into abstract interfaces.

michaelficarra: It can't be a separate PR because, once this is introduced, ModuleDeclarations are allowed when `sourceType` is "script". Too late to restrict it after that. If we don't care about strict backwards compatibility in the interface, though, it would be fine.

RReverser: And currently they're not allowed at all. We care about backwards for structures themselves, but not for abstract interfaces which just describe constraints (even now there is a lot of places like this).

Issue #6, opened 2015-02-12
Extend ExpressionStatement to indicate a directive prologue


mikesherov:
In the current syntax tree, a directive such as `"use strict"` will appear as `ExpressionStatement`. A tool that consumes the tree and needs to be aware of the strict mode will have to perform an extra step to figure out whether such an ExpressionStatement is representing a directive or not.

In this proposal, a new flag `directive` is added to `ExpressionStatement`. The value is the representation of the directive prologue in the source text.

The current state:

```
> esprima.parse('"use strict"').body[0]
{ type: 'ExpressionStatement',
  expression: 
   { type: 'Literal',
     value: 'use strict' } }
```

If this proposal is implemented, it becomes:

```
> esprima.parse('"use strict"').body[0]
{ type: 'ExpressionStatement',
  expression: 
   { type: 'Literal',
     value: 'use strict', },
  directive: 'use strict' }
```

Additional resources:
- ECMAScript 5.1 [Section 14.1](http://es5.github.io/#x14.1) on _Directive Prologues and the Use Strict Directive_
- Previous Esprima discussion: https://code.google.com/p/esprima/issues/detail?id=330.
- SpiderMonkey issue: https://bugzilla.mozilla.org/show_bug.cgi?id=791294
- See discussion on jquery/esprima#1006 for more information

michaelficarra:
A few corrections:

* the `raw` value is a non-standard esprima extension, and should not be included here
* the `directive` value must be the representation in the source code; a Boolean is not sufficient
* the `directive` value could be on either the `ExpressionStatement` or the `Literal`; they both have drawbacks

See discussion on jquery/esprima#1006 for more information.

mikesherov:
@michaelficarra corrected... thanks!

michaelficarra:
> The value is set to `true` for a directive prologue.

Should be replaced with

> The value is the representation of the directive prologue in the source text.

mikesherov:
Gah. Got it.

sebmck:
> the `raw` value is a non-standard esprima extension, and should not be included here

TIL. Should it be standardised?

mikesherov:
@sebmck file an issue?

michaelficarra:
@sebmck: What value does it have? Remember, ASTs don't have to be created by a parser.

sebmck:
@michaelficarra Yeah that's true. Can a wiki page be made on jquery/esprima for things like this?

mikesherov:
[...]

Issue #98, opened 2015-08-01
Allow null for FunctionDeclaration and ClassDeclaration id


nzakas:
`FunctionDeclaration` and `ClassDeclaration` don't allow `null` in its `id` property.
- https://github.com/estree/estree/blob/master/spec.md#functiondeclaration
- https://github.com/estree/estree/blob/master/es6.md#classdeclaration

However, both can be default exports without an ID.

``` js
espree.parse(
  "export default function() {}",
  {ecmaFeatures: {modules: true}}
);
console.log(JSON.stringify(ast, null, 4));
```

``` json
{
    "type": "Program",
    "body": [
        {
            "type": "ExportDefaultDeclaration",
            "declaration": {
                "type": "FunctionDeclaration",
                "id": null,
                "params": [],
                "body": {
                    "type": "BlockStatement",
                    "body": []
                },
                "generator": false,
                "expression": false
            }
        }
    ],
    "sourceType": "module"
}
```

I believe Acorn and Esprima produce the same. 

I think this was an oversight and the spec needs updating.

jmm:
That's interesting. Acorn / Babylon currently parse that as `FunctionExpression`, which seems to be an error.

RReverser:
No, it was intended decision (and there were debates on this) because allowing null as id in FunctionDeclaration nodes breaks backward compatibility for tools that rely on safely using id as non-null Identifier. Also, it's easier to distinguish export default declarations that can be hoisted from those that can't by simply checking whether it's expression (that is, in `export default function f()` function can be simply moved out, while in `export default 42` or `export default function ()` or `export default () => ...` it can't).

jmm:
@RReverser Interesting. (For anyone arriving at this issue, I found some of that previous discussion in https://github.com/estree/estree/pull/38, https://github.com/jquery/esprima/issues/311, https://github.com/marijnh/acorn/issues/208.)

nzakas:
That's interesting, I hadn't thought about hoistable vs. not. Still, it seems that something has to change here, either Acorn or the spec + Esprima/Espree because right now, this is an incompatibility that isn't allowed by the spec: https://github.com/estree/estree/blob/master/es6.md#exportdefaultdeclaration

@jeffmo?

jeffmo:
@rreverser mentions a debate -- which I think he and I were on opposite sides of :)

[...]

PR #99, opened 2015-08-12
Closes estree/estree#6

jasonLaster:
This picks up @dherman's [earlier commit](https://github.com/dherman/estree/commit/70ba4ab1fd8de48ca9dfd94a79820d3fe65d4009) where it left off.

Also attempted to addres @michaelficarra's concern about raw string source's being wrapped with quotes.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -37,7 +37,7 @@ interface Position {
 ```js
 interface Program <: Node {
     type: "Program";
-    body: [ Statement ];
+    body: [Directive, ..., Statement ];
 }
 ```
 
@@ -50,7 +50,7 @@ A complete program source tree.
 interface Function <: Node {
     id: Identifier | null;
     params: [ Pattern ];
-    body: BlockStatement;
+    body: FunctionBody;
 }
 ```
 
@@ -88,6 +88,16 @@ interface BlockStatement <: Statement {
 
 A block statement, i.e., a sequence of statements surrounded by braces.
 
+## FunctionBody
+
+```js
+interface FunctionBody <: BlockStatement {
+    type: "BlockStatement";
+    body: [ Directive, ..., Statement ];
+}
+```
+
+The body of a function, which is a block statement that may contain a directive prologue.
 
 ## ExpressionStatement
 
@@ -99,6 +109,18 @@ interface ExpressionStatement <: Statement {
 ```
 
 An expression statement, i.e., a statement consisting of a single expression.
+
+## Directive
+
+```js
+interface Directive <: ExpressionStatement {
+    directive: string;
+}
+```
+
+A directive from the directive prologue of a script or function.
+The `directive` property is the raw string source of the directive.
+This source includes single or double wrapping quotes and may not contain an EscapeSequence or LineContinuation..
 
 
 ## IfStatement
--- a/spec.md
+++ b/spec.md
@@ -37,7 +37,7 @@ interface Position {
 ```js
 interface Program <: Node {
     type: "Program";
-    body: [Directive, ..., Statement ];
+    body: [Directive | Statement ];
 }
 ```
 
@@ -93,11 +93,11 @@ A block statement, i.e., a sequence of statements surrounded by braces.
 ```js
 interface FunctionBody <: BlockStatement {
     type: "BlockStatement";
-    body: [ Directive, ..., Statement ];
+    body: [ Directive | Statement ];
 }
 ```
 
-The body of a function, which is a block statement that may contain a directive prologue.
+The body of a function, which is a block statement that may begin with directives.
 
 ## ExpressionStatement
 
@@ -120,7 +120,7 @@ interface Directive <: ExpressionStatement {
 
 A directive from the directive prologue of a script or function.
 The `directive` property is the raw string source of the directive.
-This source includes single or double wrapping quotes and may not contain an EscapeSequence or LineContinuation..
+This value includes single or double wrapping quotes.
 
 
 ## IfStatement


spec.md
```diff
@@ -88,6 +88,16 @@ interface BlockStatement <: Statement {
...
+## FunctionBody
+
+```js
+interface FunctionBody <: BlockStatement {
+    type: "BlockStatement";
+    body: [ Directive, ..., Statement ];
+}
+```
+
+The body of a function, which is a block statement that may contain a directive prologue.
```
gibson042: A Directive Prologue is always present (but sometimes empty). Perhaps "… a block statement that may begin with directives"?

spec.md
```diff
@@ -100,6 +110,18 @@ interface ExpressionStatement <: Statement {
...
+
+```js
+interface Directive <: ExpressionStatement {
+    directive: string;
+}
+```
+
+A directive from the directive prologue of a script or function.
+The `directive` property is the raw string source of the directive.
+This source includes single or double wrapping quotes and may not contain an EscapeSequence or LineContinuation..
```
gibson042: I don't think quotes should be included in the value. Also, `EscapeSequence` and `LineContinuation` _are_ allowed, they're just not interpreted for the purposes of testing a Use Strict Directive: http://www.ecma-international.org/ecma-262/6.0/#sec-directive-prologues-and-the-use-strict-directive .

spec.md
```diff
@@ -37,7 +37,7 @@ interface Position {
 ```js
 interface Program <: Node {
     type: "Program";
-    body: [ Statement ];
+    body: [Directive, ..., Statement ];
```
RReverser: This is not allowed by current DSL (and unclear which type `...` refers to, thus will break auto-generation of typescript and other definitions). In all the places we use smth like `[ Directive | Statement ]`).

spec.md
RReverser: Also, we will be now losing locations of directive (string literal) iself. I'd rather leave `expression` inside of `ExpressionStatement` but restrict it to `Literal`, so that new interface would remain backward-compatible.

spec.md
RReverser: Alternatively, we could put separate array `directives` in front of `body`.

RReverser: (Same for `Program` node)

gibson042: > Alternatively, we could put separate array `directives` in front of `body`.

That would either remove directives from `body` (which breaks backwards compatibility) or reference them twice (which is not JSON-friendly).

spec.md
gibson042: What do you mean? This subclasses `ExpressionStatement`, so `expression` is still present and still a `Literal`, complete with location data. It's purely additive.

spec.md
RReverser: Please see my comment about `interface Directive` below - it already breaks backward compatibility!

gibson042: How does inheriting from `ExpressionStatement` to _add_ a property without changing anything else (including `type`) break backwards compatibility?

spec.md
RReverser: I see. In that case, it violates principle `#3` from our [Philosophy](https://github.com/estree/estree#philosophy) by duplicating data that is already in `raw`.

RReverser: This seems impossible to add without breaking either one or another (and I see both as a no-go).

RReverser: What can be better instead would be adding `directive` boolean flag in this extension of `ExpressionStatement`. That would make it easy to check that statement is a directive (which is the entire point of such kind of addition) without breaking backward compatibility and at the same time data won't be duplicated and thus no problems with synchronization on node changes.

jasonLaster: ##### There are three options for the directive value:

1. the **raw value** with quotes, which preserves character information
2. the **directive value**, which makes it easy to interpret what type of directive it is
3. a **boolean value**, which prevents duplication and node syncing

###### 1. raw value
```json
{ 
 "type": "ExpressionStatement",
 "expression": { 
   "type": "Literal",
    "value": "use strict"
  },
 "directive": "\"use strict\"" 
}
```

###### 2. directive value
```json
{ 
 "type": "ExpressionStatement",
 "expression": { 
   "type": "Literal",
    "value": "use strict"
  },
 "directive": "'use strict" 
}
```

###### 3. boolean value
```json
{ 
 "type": "ExpressionStatement",
 "expression": { 
   "type": "Literal",
    "value": "use strict"
  },
 "directive": true 
}
```

I'd support the second option `directive value`. If  #14 lands, then raw data does seem redundant. Also, If directive values could be different from raw source values, then it seems helpful to flag which directive was matched.

spec.md
jasonLaster: `[Directive | Statement]` makes sense. 

My concern would be that it does not specify the order of the list. Ideally, I'd like `[Directive..., Statement...]` but I doubt that is supported.

spec.md
gibson042: > I see. In that case, it violates principle <code>#3</code> from our <a href="https://github.com/estree/estree#philosophy">Philosophy</a> by duplicating data that is already in <code>raw</code>.

`raw` is not part of this specification.

> What can be better instead would be adding <code>directive</code> boolean flag in this extension of <code>ExpressionStatement</code>.

The value must be a string instead of a boolean because otherwise `"use strict"` and `"use\x20strict"` would yield identical ASTs.

spec.md
```diff
@@ -100,6 +110,18 @@ interface ExpressionStatement <: Statement {
...
+
+```js
+interface Directive <: ExpressionStatement {
+    directive: string;
+}
+```
+
+A directive from the directive prologue of a script or function.
+The `directive` property is the raw string source of the directive.
+This value includes single or double wrapping quotes.
```
gibson042: I don't think quotes should be included in the value. `"use strict"` and `'use strict'` should yield identical ASTs.

spec.md
RReverser: @gibson042 Exactly. `raw` in fact is not needed when you have `directive` boolean flag which would simply state that 1) this `ExpressionStatement` contains string `Literal` inside of it and 2) `Literal.value` contains valid (unescaped) directive.

Why do you think it wouldn't work in this case? They wouldn't be identical since second example will not have `directive` flag (implicit `false`).

RReverser: Moreover - your comment about quotes below gets auto-resolved too since it's already solved by `value`.

gibson042: > <code>raw</code> in fact is not needed when you have <code>directive</code> boolean flag which would simply state that 1) this <code>ExpressionStatement</code> contains string <code>Literal</code> inside of it and 2) <code>Literal.value</code> contains valid (unescaped) directive.

That's not backwards compatible because `Literal.value` _is_ escaped.

> Why do you think it wouldn't work in this case? They wouldn't be identical since your second example will not have <code>directive</code> flag (implicit <code>false</code>).

No, `"use\x20strict"` can be part of a Directive Prologue per [§14.1.1](http://www.ecma-international.org/ecma-262/6.0/#sec-directive-prologues-and-the-use-strict-directive) (and therefore its `ExpressionStatement` node needs a non-null `directive` property), it's just not a Use Strict Directive.

To provide a concrete example, the following programs have three identical directives (and in particular are both strict):
```js
"use strict"; "use\x20strict"; "yustrikt";
```
```js
"yustrikt"; "use\x20strict"; "use strict";
```

but these have no directives:
```js
; "use strict";
```
```js
("use strict")
```

ESTree needs to correctly capture the distinctions, which is not possible by adding only a boolean `directive` property.

spec.md
michaelficarra: Agreed. They're insignificant.

spec.md
RReverser: > is escaped

Um, no it's not. `Literal.value` is represented unescaped (decoded) in ESTree.

RReverser: I see your point about `use\x20strict` being valid directive, but your other examples can be totally covered by boolean `directive` and I don't really see arguments why you claim it can't be. It's just a task of parser to mark valid directives with corresponding flag, and it's very easy to add.

As for escapes - having yet another CST property (`directive` string) in tree, when we have `value` and semi-standardized `raw` already. It would violate principles of this spec, and make life much harder for those who write AST transformations as they will need to synchronize all the places that represent essentially the same value in different forms (both encoded and unencoded).

This should be handled as part of #41 in generic way and it's either standardizing already-used `raw` in `Literal`s or treating CST things (like this one) as API or some other way which, again, would not violate our principles.

gibson042: > I see your point about `use\x20strict` being valid directive, but your other examples can be totally covered by boolean `directive` and I don't really see arguments why you claim it can't be.

Because the presence of escape sequences and line continuations—which are not exposed in ESTree—is relevant. String `ExpressionStatement.directive` and `ExpressionStatement.expression.value` are _not_ duplicate data; but they are obviously related (and coincide in common cases).

> having yet another CST property (`directive` string) in tree, when we have `value` and semi-standardized `raw` already.

I disagree that string `directive` is not abstract (i.e., belonging in strictly optional CST properties), because the distinctions matter for directives in a way that they don't for non-directive literals. And you can't point to `raw`, not just because it isn't part of the spec but also because _all_ properties can be derived from raw source.

RReverser: > Because the presence of escape sequences and line continuations—which are not exposed in ESTree—is relevant. 

They are not exposed in AST, but they are taken into account by parser which anyway generates those fields. Otherwise you could say the same about any ASI productions - line breaks are not exposed, but parser has to take them into account when generating AST.

RReverser: > not just because it isn't part of the spec but also because all properties can be derived from raw source

Not all of them, and, well, Literal currently has only one apart from locations, and any other node doesn't have `raw` in this meaning.

gibson042: > > Because the presence of escape sequences and line continuations—which are not exposed in ESTree—is relevant.

> They are not exposed in AST, but they are taken into account by parser which anyway generates those fields. Otherwise you could say the same about any ASI productions - line breaks are not exposed, but parser has to take them into account when generating AST.

It is a mistake to assume that ASTs are only generated by parsers, but beside the point anyway—this is about ensuring that semantically different programs don't map to identical ASTs.

RReverser: For non-parsers none of the cases simplifies generation, as both add complexity to AST (but, well, it's evolution). As for semantics - we are not trying to reflect semantics in AST at all, we are reflecting syntax (it's Syntax Tree after all) and thus in many cases map semantically different pieces of programs to identical ASTs.

sebmck: Just standardise raw. Once that's done this won't be necessary. Parent nodes should not be concerned about what their children represent (and vice versa), it's a part of the ESTree philisophy.

RReverser: @sebmck That's exactly what I wrote above. Thank you!

sebmck: @RReverser Yup, just backing you up!

gibson042: > many cases map semantically different pieces of programs to identical ASTs

Each such case would be a spec bug. However, I'm unaware of any not already captured as issues (e.g., #6 and #98), and those hardly count as "many cases". Could you provide some examples?

> Just standardise raw.

I'm in favor of an optional—it must be optional to maintain backwards compatibility—`Literal#raw`, but note that it does come with the very same data duplication problems being discussed here ("[AST transformations… will need to synchronize all the places that represent essentially the same value in different forms](https://github.com/estree/estree/pull/99#discussion_r36993683)"). _And_ we would **still** need to introduce a `directive` property (e.g., for differentiating `"use strict"` from the parenthesized and therefore directive-free `("use strict")`), but it could be boolean.

So if we go that route (adding boolean-bearing `Directive <: ExpressionStatement` and string `Literal#raw`), code to check for strict mode would look like
```js
function isStrict( statementList ) {
  for ( let i = 0, statement, directive; (statement = statementList[i]); i++ ) {

    // End of Directive Prologue?
    if ( !statement.directive ) { break; }

    // Get raw string, defaulting to backwards-compatible `value`
    // (assuming that `directive` ⇒ `raw != null || raw.slice(1, -1) === value`)
    directive = (statement.expression.raw ||
        ("'" + statement.expression.value + "'")).slice(1, -1);

    // Require exact match
    if ( directive === "use strict" ) { return true; }
  }

  // No "use strict" directive
  return false;
}
```

... which has a more complicated body than the alternative
```js
    if ( !statement.directive ) { break; }
    if ( statement.directive === "use strict" ) { return true; }
```
but moves us closer to resolving #41, and gracefully condenses EscapeSequence- and LineContinuation-free string Literals wherever they appear.

RReverser: `value` should be never trusted as directive source due to escaping, so if it's not present, tool can just `return false` immediately. If node generator (whether parser or custom builder) wants to generate directive, it will be just obliged to put `raw` into it (same as, for example, with generating Literal nodes for asm.js which distinguishes `2.0` from `2`).

This way, first example will be simplified to:

```javascript
return statement.directive && (statement.expression.raw || '').slice(1, -1) === "use strict";
```

Which is not complicated and in any case you just write it once and then reuse.

P.S. `raw` is already present in nodes generated by all the existing parsers, so it's not breaking backwards compatibility but standardizing it (as we did with most "non-standard" features when ESTree was initiated).

RReverser: In fact, @sebmck is right - making parent to know anything about it's children is really, really bad even if it's just a boolean flag which depends on type of child node. Instead, we chould solve CST in general and provide way to distinguish parenthesized expressions from regular ones.

sebmck: Yeah, to add onto what @RReverser is saying. You can't trust `directive` on an `ExpressionStatement` if there's even the possibility of transforming it. Whenever you wanted to replace an expression you'd have to check if the parent was an `ExpressionStatement` and invalidate it if it's a directive which is rather unnecessary.

RReverser: What I think is that more reliable solution could be to have some `parenthesized` flag on `Expression` iself (which we optionally in Acorn with corresponding config, but obviously not standardized right now).

gibson042: > P.S. `raw` is already present in nodes generated by all the existing parsers, so it's not breaking backwards compatibility but standardizing it (as we did with most "non-standard" features when ESTree was initiated).

Again I say, parsers are not the only source of ASTs. Do you really want to make `raw` required, even if only for directives?

> we chould solve CST in general

Agreed, but my impression was that full CST components will be optional, so they can't be relied upon for directive testing.

> have some parenthesized flag on Expression iself

It's true, AFAICT `directive` is not required if we have both `Expression#parenthesized` and `Literal#raw`.

RReverser: > Do you really want to make raw required, even if only for directives?

It's not just for directives, and I do know that parsers are not the only source - that's why I wrote about adding it by tools.

> It's true, AFAICT directive is not required if we have both Expression#parenthesized and Literal#raw.

Yup, that's what I'm thinking about as more proper solution that wouldn't break any of our rules.

gibson042: > > Do you really want to make raw required, even if only for directives?

> It's not just for directives, and I do know that parsers are not the only source - that's why I wrote about adding it by tools.

My point is that the spec already exists; _requiring_ `raw` outside of a new node is not backwards compatible. _Optional_ `raw` is compatible, though, and simpler to produce—but slightly more complex to consume (e.g., `literal.raw || ("'" + literal.value + "'")`).

RReverser: This spec is being developed in two ways: first one is adding new things in the form that every member listed in README agrees on; second one is adding old things that every member listed in README already implements / depends on. `raw` is example of second one.

RReverser: > but slightly more complex to consume (e.g., `literal.raw || ("'" + literal.value + "'")`)

No, `value` should never be used as fallback for `raw`. Please see my comments above about this and which way would be proper and work even with optional `raw` that you suggest.

michaelficarra: @gibson042 It's not that simple either: a directive can have an unescaped `'` if it was defined using `"` as the delimiter.

gibson042: Yes, all understood. My assumption was that an optional `raw` would allow producers to omit the property when it is directly implied by value (i.e., a value for which `toString` is idempotent and outputs no line terminators or potential delimiters), reducing data duplication. Such looseness could easily be prohibited, though—even on a case-by-case basis in consumers, like @RReverser's suggestion that absence of `raw` implies absence of directives.

PR #100, opened 2015-09-18
ImportSpecifier: add local field

hzoo:
It's referenced the the explanatory paragraph below, but should it also go here as well?


btmills:
`local` is already inherited from [`ModuleSpecifier`](https://github.com/estree/estree/blob/master/es6.md#modulespecifier).

hzoo:
Ok cool thanks @btmills

Issue #66, opened 2015-03-19
Add example documents


twolfson:
A while ago when I was working on `esformatter-phonetic`, I was frustrated at the lack of examples in the SpiderMonkey API documentation. To resolve my frustration and potential for error, I wound up creating `ecma-scopes` to programmatically determine when a scope is lexical/block level.

https://github.com/twolfson/esformatter-phonetic

https://github.com/twolfson/ecma-variable-scope

https://github.com/twolfson/ecma-scopes

I considered going all the way and writing a full suite for every basic case as examples for future developers. However, realized how large a mountain that was. With `estree`, it seems like a great place to keep examples along side core documentation. Would you be interested in adding such examples?

One example of the proposed examples would be:

# Identifier
A name for a reference or property of an object. Given the following example:

```js
var foo = 'bar';
function baz(xyz) { }
foo.wat;
foo.huh();
```

Each of `foo`, `baz`, `xyz`, `wat`, and `huh` are considered identifiers. An example of the AST representation is:

```js
{
  type: 'Identifier',
  name: 'foo'
}
```

impinball:
Might not be a bad idea, although you probably meant for your example to
look like this:

```json
{
  type: "Identifier",
  name: "foo"
}
```

mikesherov:
@twolfson would love some examples here. PRs welcome.

twolfson:
Unfortunately, my plate is much more full now than it was a few months ago. I doubt I will be able to work on this any time soon =(

hzoo:
Is @twolfson's example in the OP what we would want? I can work on this

PR #105, opened 2015-10-16
WIP: add examples

hzoo:
Just to get the idea rolling

Ref https://github.com/estree/estree/issues/66

@twolfson 's example

## Identifier

A name for a reference or property of an object. Given the following example:
```js
var foo = 'bar';
function baz(xyz) { }
foo.wat;
foo.huh();
```
Each of `foo`, `baz`, `xyz`, `wat`, and `huh` are considered identifiers. An example of the AST representation is:
```js
{
  type: 'Identifier',
  name: 'foo'
}
```

Not sure what we want - seems like it could go in a lot of directions? Should this be in another folder and linked to the other files?

Do we want to have a short description of what each node type is used for (what I put is placeholder)? And what part of the AST do we what to be showing? Do we want to show all the different kinds of syntax (in this case both FunctionExpression and FunctionDeclaration)?

### DIFF
--- a/examples/es6.md
+++ b/examples/es6.md
@@ -1,0 +1,61 @@
+This document specifies the extensions to the core ESTree AST types to support the ES6 grammar.
+
+# Functions
+
+You can have generator function declarations
+
+```js
+function * a() {}
+```
+
+```js
+{
+    "type": "FunctionDeclaration",
+    "id": {
+        "type": "Identifier",
+        "name": "a"
+    },
+    "generator": true,
+    "expression": false,
+    "async": false,
+    "params": [],
+    "body": {
+        "type": "BlockStatement",
+        "body": []
+    }
+}
+```
+
+as well as generator function expressions
+
+```js
+var a = function * () {}
+```
+
+```js
+{
+  "type": "VariableDeclaration",
+  "declarations": [
+    {
+      "type": "VariableDeclarator",
+      "id": {
+        "type": "Identifier",
+        "name": "a"
+      },
+      "init": {
+        "type": "FunctionExpression",
+        "id": null,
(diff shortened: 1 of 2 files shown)


examples/es6.md
```diff
@@ -0,0 +1,61 @@
...
+      "type": "VariableDeclarator",
+      "id": {
+        "type": "Identifier",
+        "name": "a"
+      },
+      "init": {
+        "type": "FunctionExpression",
+        "id": null,
+        "generator": true,
+        "expression": false,
```
twolfson: Is this (`expression`) really `false` in the current AST scheme?

hzoo: Hmm actually yeah it is

twolfson: Weiiiird

hzoo: Oh https://github.com/estree/estree/blob/bc81a9a8902ca25cf496a626e911f279af45ed4e/deprecated.md#functions

twolfson:
I think this is a good start. I won't comment on folder structure (I think that belongs to the repo owners).

iirc my thoughts were to keep items structred by node type (similar to `spec.md`/SpiderMonkey AST doc). Within those sections, we could provide line by line examples of their scenarios (I think `Identifier` was one of those that stood out with all its scenarios; `BlockStatement` will likely be similar due to its autoinjection in parsing things like `if's` even with no braces).

twolfson:
Although, when we look at ES5 vs ES6 -- this probably would fall apart since `FunctionDeclaration` has already been stated in the other doc =/

examples/es6.md
twolfson: Ah, interesting. Nice find =)

PR #106, opened 2015-10-31
Automate TOC updates

gibson042:
`npm run render` → `doctoc *.md`

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "estree",
   "version": "1.0.0-pre",
   "description": "ECMAScript syntax representation format",
+  "private": true,
   "dependencies": {},
   "devDependencies": {
     "doctoc": "^0.15.0"
--- a/package.json
+++ b/package.json
@@ -2,14 +2,12 @@
   "name": "estree",
   "version": "1.0.0-pre",
   "description": "ECMAScript syntax representation format",
-  "main": "spec.md",
   "dependencies": {},
   "devDependencies": {
     "doctoc": "^0.15.0"
   },
   "scripts": {
     "render": "doctoc *.md",
-    "start": "npm run render"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,0 +1,35 @@
+{
+  "name": "estree",
+  "version": "1.0.0-pre",
+  "description": "ECMAScript syntax representation format",
+  "main": "spec.md",
+  "dependencies": {},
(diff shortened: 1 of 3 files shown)


mikesherov:
:+1:

RReverser:
I specifically didn't add it for README.md as it's very small and TOC adds much overhead given that we have only couple of items.

package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "estree",
+  "version": "1.0.0-pre",
```
RReverser: Maybe makes sense to use ECMAScript edition as major version?

package.json
```diff
@@ -0,0 +1,35 @@
...
+  "description": "ECMAScript syntax representation format",
+  "main": "spec.md",
```
RReverser: Node.js uses `mail` for `require` calls, Markdown in `main` wouldn't make sense and just confuses IMO.

package.json
gibson042: The field is explicitly semver. And since this effort is committed to 100% backwards compatibility, that means the major version will never change.  We could try using minor for that purpose, but it would break down as soon as we added something in between ECMAScript releases. I think it's better to avoid cleverness here.

package.json
gibson042: Good point; I'll just omit it completely.

gibson042:
I tried to make the README TOC as unobtrusive as possible, but wanted it to avoid complexity in the npm script. But if everyone else thinks it should stay out, I can make that happen.

RReverser:
@gibson042 Sorry for the long delay here. Could you please add `"private": true` to `package.json` as we don't really plan to publish Markdown as a package to npm? And I'll happily merge.

gibson042:
@RReverser done.

RReverser:
> I tried to make the README TOC as unobtrusive as possible, but wanted it to avoid complexity in the npm script. But if everyone else thinks it should stay out, I can make that happen.

@mikesherov thoughts?

mikesherov:
@rreverser I agree with you here. npm script complexity doesn't bother me here.

gibson042:
So you don't like the [README TOC](https://github.com/gibson042/estree/blob/doctoc/README.md)?

RReverser:
@gibson042 I believe it's redundant and only increases visual size of README without having useful function.

adrianheine:
Closing per @RReverser (I agree).

PR #107, opened 2015-10-31
Define CST elements

gibson042:
Fixes gh-41

Based on https://github.com/estree/estree/issues/41#issuecomment-131636382 , with slight modifications.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1,6 +1,6 @@
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
-This document specifies the core ESTree AST node types that support the ES5 grammar.
+This document specifies the core ESTree AST and CST node types that support the ES5 grammar.
 
 - [Node objects](#node-objects)
 - [Identifier](#identifier)
@@ -60,6 +60,10 @@ This document specifies the core ESTree AST node types that support the ES5 gram
   - [NewExpression](#newexpression)
   - [SequenceExpression](#sequenceexpression)
 - [Patterns](#patterns)
+- [Concrete syntax](#concrete-syntax)
+  - [Child-node reference](#child-node-reference)
+  - [Token](#token)
+  - [Nontoken](#nontoken)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -692,3 +696,52 @@ Destructuring binding and assignment are not part of ES6, but all binding positi
 ```js
 interface Pattern <: Node { }
 ```
+
+# Concrete syntax
+
+Whitespace, comments, and the exact character sequences comprising Literal nodes are part of the lexical grammar, but not part of the _abstract_ syntax. Their representation is completely optional.
+
+```js
+interface ConcreteNode <: Node {
+    sourceElements: [ ChildReference | Token | Nontoken ];
+}
+```
+
+The `sourceElements` field is a list of input elements and references to child nodes. Any (sub)tree in which `sourceElements` is complete on every node (i.e., there are no unreferenced child nodes and all necessary whitespace and punctuation is included) can be rendered by depth-first concatenation (with respect to references).
+
+## Child-node reference
+
+```js
+interface ChildReference {
+    reference: string;
+}
+```
+
+`reference` names the property identifying a referenced child (e.g., `expression` for ChildReferences on ExpressionStatement nodes), optionally followed by `#next` when identifying the next unreferenced member of a list-valued property (e.g., `body#next` for ChildReferences on BlockStatement nodes).
+
+## Token
+
+```js
+interface Token <: ChildReference {
+    element: string;
+    reference: string | null;
+    value: string | null;
+}
+```
+
+Tokens are relevant to abstract syntax. The `element` field is a string identifying input element type ("Keyword", "Identifier", "Punctuator", "StringLiteral", etc.).
+
+The `reference` field is optional (being useful only when there is already a field matching the input token, e.g. `operator` on BinaryExpression nodes), and if absent must be replaced by a `value` of the exact character(s) corresponding to the token (e.g., `function`, `(`, `0xCAFE`).
+
+## Nontoken
+
+```js
+interface Nontoken {
+    element: string;
+    value: string;
+}
+```
+
+Nontokens are not relevant to abstract syntax. The `element` field is a string identifying type (one of "WhiteSpace", "LineTerminator", "CommentHead", "CommentBody", "CommentTail").
+
+The `value` field is identical to the same field on Nontoken elements—a string of the exact character(s) corresponding to it.


forivall:
For the sake of being able to more easily tell the type of a source element, it would probably be helpful to include `type: "ChildReference"`, `type: "Token"`, `type: "NonToken"` for each of the new types.

gibson042:
I thought about that, but `type` would collide with the Node property. Honestly, though, there's no relevant ECMAScript-level distinction. The key detail in _this_ representation is `reference` vs. `value`, which is obvious on any given instance.

spec.md
```diff
@@ -692,3 +696,52 @@ Destructuring binding and assignment are not part of ES6, but all binding positi
...
 interface Pattern <: Node { }
 ```
+
+# Concrete syntax
+
+Whitespace, comments, and the exact character sequences comprising Literal nodes are part of the lexical grammar, but not part of the _abstract_ syntax. Their representation is completely optional.
+
+```js
+interface ConcreteNode <: Node {
+    sourceElements: [ ChildReference | Token | Nontoken ];
```
nzakas: Can `ConcreteNode` show up in `sourceElements`? If not, should it?

spec.md
```diff
@@ -692,3 +696,52 @@ Destructuring binding and assignment are not part of ES6, but all binding positi
...
+
+## Child-node reference
+
+```js
+interface ChildReference {
+    reference: string;
+}
+```
+
+`reference` names the property identifying a referenced child (e.g., `expression` for ChildReferences on ExpressionStatement nodes), optionally followed by `#next` when identifying the next unreferenced member of a list-valued property (e.g., `body#next` for ChildReferences on BlockStatement nodes).
```
nzakas: This seems like a lot of overhead to identify the node. Why not have a direct reference?

spec.md
gibson042: No and no, because `ConcreteNode` extends `Node` (which corresponds to a [syntactic production](http://www.ecma-international.org/ecma-262/6.0/#sec-syntactic-grammar)) while `sourceElements` contains lexical [input elements](http://www.ecma-international.org/ecma-262/6.0/#sec-lexical-and-regexp-grammars), which are _not_ nodes.

spec.md
gibson042: Because that would not be expressible in JSON, and therefore not useful for data interchange.

spec.md
gibson042: I did it this way because I don't think adding `sourceElements: [ ChildReference | Token | Nontoken ] | null` to `Node` would be valid syntax, but I see that I missed noting that everything else should extend not `Node` but `ConcreteNode`. I'll update.

spec.md
nzakas: Gotcha.

nzakas: Looking through this again, it seems like it would be problematic to generate this during a single parser pass. Ideally, you'd want to keep track of tokens, comments, etc., as they're being parsed and just build up the list. However, you may not know what type of parent a reference will be attached to until  after all source elements are known. That would mean going through the source elements and updating the reference location later. Do you have thoughts about this?

gibson042: Backtracking _might_ be necessary if you want tight control over comment/whitespace attachment¹ (both in the before/after sense and in the depth sense), but it is definitely possible to do this in a single pass. I put together a proof of concept with off-the-shelf `acorn`: http://jsfiddle.net/k2ndj82p/

Note that it abuses `type` as a development shortcut, and uses artificial ids for references instead of the above query language—therefore also missing out on puncutator references. Fixing that would involve leveraging node type awareness when setting `sourceElements`.

¹ See `nodeFinisher` in the POC

gibson042:
Thanks for the [use cases](https://github.com/estree/estree/issues/41#issuecomment-153774133), @nzakas; they're great! I'll give them a first-pass attempt, but understand that some sharp edges are explicitly avoided, and others I'm sure I just missed.

> # Given a node, determine if it is surrounded by parentheses.

This one has some interesting properties; I'll come back to it.

> # Insert a new argument between two existing arguments of a CallExpression.

```js
/* Insert `newArg` between `a` and `b` in `node.arguments`, where `b` is at `argIndex` */
node.arguments.splice(argIndex, 0, newArg);
let srcIndex;
let fill = node.sourceElements.reduce(function(fill, input, i) {
	// Found an argument?
	if ( input.reference === "arguments#next" ) {
		// Record it by decrementing `argIndex`, and set `srcIndex` if we found `b`
		if ( argIndex-- === 0 ) srcIndex = i;

	// Copy everything between `a` and `b` except comments
	} else if ( argIndex === 0 && input.element.slice(0, 7) !== "Comment" ) {
		fill.push( Object.assign({}, input) );
	}

	return fill;
}, []);
node.sourceElements.splice(srcIndex, 0, { reference: "arguments#next" }, ...fill);
```

> # Comment out a function.

I'll assume this is a function declaration, since replacing a function expression with a comment requires more information about the context (but is otherwise similar).
```js
/* Replace the function declaration at index `i` in `node.body` with a comment */
srcIndex = indexOfRef(node.sourceElements, i); // uses the logic from above
node.sourceElements.splice(srcIndex, 1,
	createBlockCommentHead(),

	createCommentBody(
		// Escape `*/` sequences without touching existing escapes
		render(node.body[i]).replace(/\\([\w\W])|(\*)(\/)/g, "$2\\$1$3")
	),
	createBlockCommentTail()
);
node.body.splice(i, 1);
```

> # Combine multiple var statements into a single var statement.

For simplicity, I'll avoid adding indentation and assume the same `kind` for all declarations.
```js
/* Roll variable declarations into `decl`, at index `i` in `node.body` */
let declIndex = indexOfRef(node.sourceElements, i); // uses the logic from above
do {
	// Break out if there is no further declaration
	next = node.body[++i];
	if ( !next || next.type !== "VariableDeclaration" ) break;

	// Ensure the necessary comma
	let input, terminatorIndex = decl.sourceElements.length;
	while ( (input = decl.sourceElements[--terminatorIndex]) &&
		input.value !== ";" && !input.reference );
	if ( input.reference ) {
		decl.sourceElements.splice(terminatorIndex + 1, 0,
			(input={ element: "Puncutator" }));
	}
	input.value = ",";

	// Claim intervening source elements
	let nextIndex = declIndex;
	while ( (input = node.sourceElements[++nextIndex]) &&
		input.reference !== "body#next" );
	decl.sourceElements.push(
		// Pluck (but do not import) the reference to `next`
		...node.sourceElements.splice(declIndex + 1, nextIndex).slice(0, -1)
	);

	// Claim _subordinate_ source elements (including declarator references)
	decl.sourceElements.push(
		// ...but excluding the `var`/`let`/`const` keyword
		...next.sourceElements.filter( input => input.element !== "Keyword" )
	);

	// Update the AST, removing `next` and moving its declarators to `decl`
	node.body.splice(i, 1);
	decl.declarations.push(...next.declarations);
} while ( true );
```

Ok, back to the first:
> # Given a node, determine if it is surrounded by parentheses.

I think this requires access to parent nodes, but we **absolutely** prohibit the multiple (and in fact circular) references necessary to get them directly from the ESTree data structure. However, programs _processing_ trees are free to do whatever they need on input/output, be they attaching ids like [my POC](http://jsfiddle.net/k2ndj82p/), using on-the-side (Weak)Maps, or even introducing outright cycles like [CST](https://github.com/cst/cst). So, assuming some helper functions,
```js
let reNonToken = /^WhiteSpace|^LineTerminator|^Comment/;
function isParenthetical( node ) {
	// Check inwards if the first token is an open parenthesis
	let inwardsParenthetical = node.sourceElements.reduce(function(answer, input) {
		if ( answer != null ) return answer;
		if ( reNonToken.test(input.element) ) return answer;
		if ( input.value === "(" ) return true;
		return false;
	}, null);
	if ( inwardsParenthetical ) return true;

	// Check upwards
	for ( let dir of [-1, 1] ) {
		let expectedValue = dir === -1 ? "(" : ")";
		let categorize = dir === -1 ? $categorizeOpenParen : $categorizeCloseParen;
		let nut = node, parent; // "node under test"

		while ( (parent = $parentNode(nut)) ) {
			// Find the first preceding/following token
			let input, nutIndex = $sourceIndexOf(nut, parent);
			while ( (input = parent.sourceElements[nutIndex += dir]) &&
				reNonToken.test(input.element) );
			if ( input ) {
				// If it's not a grouping parenthesis, we know all we need to
				if ( input.value !== expectedValue ||
					categorize(input, parent) !== "ParenthesizedExpression" ) {

					return false;
				}

				// Otherwise, we may need to check the other side
				continue;
			}
			nut = parent;
		}

		// If we run out of "up", `node` is not parenthesized
		return false;
	}

	// We `continue`d twice, so `node` _is_ parenthesized
	return true;
}
```

forivall:
rough implementation at https://npmjs.com/cstify; test at http://forivall.com/astexplorer/

PR #110, opened 2015-12-07
Eliminate redundant Node inheritance references.

jmm:
Isn't `Node` redundant here since these extend `Expression`, which extends `Node`? The various flavors of `Statement` just extend `Statement`.


RReverser:
Fair enough.

PR #111, opened 2015-12-08
Include Identifier as possible type of MemberExpression.property

jmm:
I think this is a more preferable way to define the interface? Or, if you're comfortable with it, https://github.com/estree/estree/compare/master...jmm:member-expression-property2.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -632,7 +632,7 @@ A logical operator token.
 interface MemberExpression <: Expression, Pattern {
     type: "MemberExpression";
     object: Expression;
-    property: Expression;
+    property: Expression | Identifier;
     computed: boolean;
 }
 ```


RReverser:
`Identifier` is already a subset of `Expression`, so it doesn't make sense.

jmm:
Ah, good point. I should've realized that after my PR yesterday...d'oh! Sorry for the noise. Unless you like the ternary-like representation, `computed ? Expression : Identifier`, this can be closed. Thanks @RReverser.

On another note, perhaps it'd make sense to say:

> If `computed` is `true`, the node corresponds to a computed (`a[b]`) member expression and `property` is an `Expression` (possibly an `Identifier`).

Especially since in the example that will be the case. But technically it's redundant like what I suggested in this PR.

RReverser:
Ternary is not really a part of the DSL and I'm not sure we want to extend it significantly. As for docs, yes, such wording might work, please open another PR if you wish.

PR #87, opened 2015-05-31
Added Exponentiation Operator

nicolo-ribaudo:
Proposal: https://github.com/rwaldron/exponentiation-operator

PR #116, opened 2016-01-21
Create experimental/decorators.md

hzoo:
babel ast: http://astexplorer.net/#/0fkNiWEMLp

### DIFF
--- a/experimental/decorators.md
+++ b/experimental/decorators.md
@@ -1,0 +1,26 @@
+# [Decorators](https://github.com/wycats/javascript-decorators)
+
+## Decorator
+
+```js
+interface Decorator <: Node {
+    type: "Decorator";
+    expression: Expression;
+}
+```
+
+## Used
+
+```js
+extend interface MethodDefinition {
+    decorators: [ Decorator ];
+}
+
+extend interface Property {
+    decorators: [ Decorator ];
+}
+
+extend interface Class {
+    decorators: [ Decorator ];
+}
+```


forivall:
`ClassDeclaration` and `ClassExpression` will also need a decorators prop.

hzoo:
Thanks added

experimental/decorators.md
```diff
@@ -0,0 +1,26 @@
...
+interface Decorator <: Node {
+    type: "Decorator";
+    expression: Expression;
+}
+```
+
+## Used
+
+```js
+extend interface MethodDefinition <: Node {
```
RReverser: This is an extension of existing `interface`, so it can't have parent specifier `<: Node`.

hzoo: Oh right - should it just be `extend interface MethodDefinition {` then?

RReverser: Yup. You can see `es6.md` for examples.

hzoo: Ah ok, I'l need to fix async/await then as well?

hzoo:
Ok removed

experimental/decorators.md
RReverser: Same problem then, didn't notice earlier.

RReverser:
LGTM, please squash and I'll merge.

hzoo:
Wish you could squash in github :cry: k.

RReverser:
Yeah :( And rebase instead of merge.

hzoo:
Ok done

nzakas:
Hey all, I'd just like to raise a flag here. It seemed like there was little (no?) discussion around this before being merged (and only 11 hours before merging, not enough time for different time zones to weigh in). While I'm all for moving quickly, this repo was setup to get consensus on things, so we all agreed on the right approach. I don't see any consensus building around this change, and I'd like to suggest that going forward things not move so quickly so others have time and opportunity to respond.

RReverser:
@nzakas Hmm, this is experimental folder for non-standardized things that are just implemented in few implementations. Do you think adding them requires discussion, too?

hzoo:
Sorry about that, I figured since it's experimental/can change at anytime I could add some PRs of what's currently implemented.

RReverser:
Same to what @hzoo said :/

nzakas:
Yes, otherwise we are just documenting other ASTs, and that doesn't make much sense. Whatever ends up in this repo should have some sort of growing consensus.

Babel can always document its own stuff.

nzakas:
What's the harm in talking about it first, anyway? Maybe everyone will agree that these are the way to go? What's the hurry?

RReverser:
@nzakas Babel uses completely different AST anyway since version 6, this is unrelated. Experimental always was just a playground, where we add/remove things. And if timing is an argument here, then I have to note that there are experimental PRs that stay around for almost a year like https://github.com/estree/estree/pull/87 without any discussion happening at all.

mikesherov:
There should be at least some consensus. A helpful ping of myself or @ariya and @nzakas etc. would be nice. Can always say, "merging in a few days if there's no disagreement"

RReverser:
@mikesherov Does Esprima implement any experimental features? I thought it, just like Acorn, waits for new standard to be adopted. In that case, my understanding was that everyone is welcome to provide some initial ideas for how AST will look like in experimental folder, and when it becomes a standard, we clean things up and move to actual spec.

RReverser:
Previous PRs to `experimental` that led me to this assumption:
https://github.com/estree/estree/pull/26 (this one contained "merging in a few days" statement, but inacitivity != agreement IMO, and there was still no discussion)
https://github.com/estree/estree/pull/103 (no discussion)

mikesherov:
No worries about assumption. 

I'm torn here. I would think having another set of eyes on it is good just from contributor experience level. Also, want to make sure we have some input so when experimental becomes real, there's already some buy in with hopefully little churn.

On the other hand, experimental *is* experimental and I personally feel comfortable with the direction here.

All in all, I suppose there's no harm in pinging and waiting a day or two before merging. Is that OK?

RReverser:
Agreed, will wait in the future.

RReverser:
Given that, since we touched this topic, I would ask participating contributors to watch the repo and [since I know how many notifications you get daily and how easy it's to miss some] to visit the repo from time to time and see whether smth needs approval / discussion. Otherwise, I'm afraid, external visitors might think that spec is dead when nothing happens on issues/PRs for weeks or even months like seen in some cases.

mikesherov:
Fair enough. 

Mike Sherov

PR #118, opened 2016-01-21
UnaryExpression are always prefix

amasad:
The `prefix` boolean field on UnaryExpressions is redundant because all operators are currently prefix.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -508,7 +508,6 @@ A `function` expression.
 interface UnaryExpression <: Expression {
     type: "UnaryExpression";
     operator: UnaryOperator;
-    prefix: boolean;
     argument: Expression;
 }
 ```


ariya:
While it's likely harmless, doesn't this fall into "Non-additive modifications"?

forivall:
I'm speculating that it's so that `UpdateExpression` and `UnaryExpression`s can be handled by the same functions, or some other historical reason like a Mozilla js extension that added non-prefix unary expressions.

RReverser:
Yeah, we discussed this with @dherman and, unfortunately, this is one of redundant design decisions that were here from the beginning and thus breaking to remove.

amasad:
Ok, makes sense.

PR #122, opened 2016-01-27
Add type annotations (fixes #115)

### LINKED ISSUE #115: Proposal for open-ended type annotations
With things like TypeScript and Flow becoming more popular, I'd really like to find a way to incorporate type annotations into ESTree in some sort of common way so we can start creating tooling around it. In particular, we're getting more requests in ESLint to be able to write rules around type annotations, but without having this is information in a predictable place, it's really hard to do that effectively.

I've thrown together a strawman below based on looking at what the TypeScript compiler produces and what Babel currently produces (neither works directly for ESTree, IMHO). I'm not tied to any of this, just want to start the conversation.

**Note:** What I'm proposing here is not necessarily TypeScript or Flow extensions to ESTree, just extensions that allow for type annotations. None of this relates to defining types, interfaces, etc.

## TypeAnnotation

First, suppose there's an abstract interface called `TypeAnnotation` from which TypeScript and Flow types can inherit as needed:

```js
interface TypeAnnotation <: Node {}
```

This is really just a placeholder in the ESTree format.

## TypedPattern

It's a little bit tricky to handle named function arguments with a type, such as `function x(a:string)` because we use `Identifier` (or `AssignmentPattern` for params with defaults). It doesn't seem correct to add something to `Identifier`, as identifiers don't generally have type annotations, and that also wouldn't handle the other cases. So, I think the easiest thing to do is to introduce a new node that represents a name-type pair.

```js
interface TypedPattern <: Pattern {
    id: Identifier;
    dataType: TypeAnnotation;
}
```

So in the case of `function x(a:string)`, the argument would be represented by a `TypedPattern` node. When there's a default value, such as `function x(a:string="")`, then the argument is an `AssignmentPattern` whose `left` property is a `TypedPattern`. 

There would be no need to extend the interface of `FunctionDeclaration`, `FunctionExpression`, or `ArrowFunctionExpression` since all of these use `Pattern` to define their parameters.

## Functions

In order to support a return value for functions (for example `function x(a:string):string {}`), we do need to extend the various function nodes:

```js
extend interface FunctionDeclaration {
    returnType: TypeAnnotation | null;
}

extend interface FunctionExpression {
    returnType: TypeAnnotation | null;
}

extend interface ArrowFunctionExpression {
    returnType: TypeAnnotation | null;
}
```

## VariableDeclarator

To support variable declarations, such as `var foo:string = "bar"`, we can simply used a `TypedPattern` in as the `VariableDeclarator`'s `id` property. The `id` property is already defined as `Pattern`, so no change to the `VariableDeclarator` interface is necessary.

## Conclusion

I think there's a fairly lightweight way to support type annotations formally in ESTree without typing us down to a particular flavor. As I said, I'm not typed to this proposal, I'd really just like to get some consensus as I think it could improve the state of typed-JavaScript tooling going forward.

nzakas:


RReverser:
I guess this goes to `experimental` as it's not official ECMAScript? Also, doesn't cover type casts at the moment.

nzakas:
What are type casts in this context?

nzakas:
maybe not an `extensions` folder instead of `experimental`?

RReverser:
> What are type casts in this context?

@nzakas Flow: `(obj.method(): SomeType)`, TypeScript: `<SomeType>obj.method()`

Also interface declarations, class implements, module declarations, type aliases etc. (most of this can be mapped 1:1 from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/estree/flow.d.ts).

Or do you want just to add generic type annotations and only on identifiers?

nzakas:
Ah yeah, as I said in the original issue:

> Note: What I'm proposing here is not necessarily TypeScript or Flow extensions to ESTree, just extensions that allow for type annotations. None of this relates to defining types, interfaces, etc.

Just want to know where annotations go when declaring variables and functions.

DanielRosenwasser:
TypeScript also has the `obj.method() as SomeType` syntax.

DanielRosenwasser:
should there also be annotations on `ObjectPattern` and `ArrayPattern`?

nzakas:
@DanielRosenwasser I don't know, you tell me.

DanielRosenwasser:
If the plan is to reflect the current way Flow does it, then yeah. I'd say just put it in *Pattern*, but I don't see it defined anywhere.

nzakas:
@jeffmo thoughts?

jeffmo:
That's right -- currently `typeAnnotation` can sit on the various patterns to satisfy things like `var {a}: {a: number} = foo;}`

(see also: https://astexplorer.net/#/wTQEak360w/2)

nzakas:
So `typeAnnotation` would be on `Pattern`? Or better to have it on `ObjectPattern` and `ArrayPattern`, specifically?

RReverser:
> I'd say just put it in Pattern

This would affect `MemberExpression` and `AssignmentPattern` which don't have annotations.

I would put on `{Object,Array}Pattern` and `RestElement` manually.

jeffmo:
What @RReverser said :)

nzakas:
Sounds good.

nzakas:
Updated.

nzakas:
@RReverser @jeffmo @mikesherov @ariya anything preventing this from being merged now?

RReverser:
Looks fine to me (smth we can work with and extend in future). Waiting for feedback from others.

ariya:
Unfortunately I haven't put too much thought on type annotation :crying_cat_face:

RReverser:
@ariya Well, this goes into just an experimental folder (so not something ESTree implementors are obliged to implement, but rather "formalized thoughts on the subject").

jeffmo:
This seems simple enough to be flexible for many common circumstances. A few "common" things it skips (intentionally I believe -- just pointing them out for posterity): type parameter lists on functions/classes, type aliases (also w type params), interfaces (w/ type params), and then of course describing a format for the typeAnnotation node.

Overall seems like a good start to me

nzakas:
@jeffmo re: type parameter lists. It seems like that's not needed because type annotations are directly on the parameters?

You are correct that the rest are intentionally omitted and left to individual dialects for now.

jeffmo:
> It seems like that's not needed because type annotations are directly on the parameters?

By type parameters I mean things like `T` in `function foo<T>() { ... }` or `class Foo<T> extends Bar<T> {}`. For this you'd want something like:

```js
interface TypeParameter {
  type: "TypeParameter",
  name: Identifier
}

extend interface Function {
  typeParameters: [ TypeParameter ];
  returnType: TypeAnnotation | null;
}
```

(Note that Flow used to lack an intermediary `TypeParameter` node and just use `Identifier`, but this turned out to be a bad call for various reasons that I could go into, but I suspect might be out of scope of this proposal if it omits type parameters)

paulyoung:
Babylon already includes an AST format for type annotations so @kittens might have valuable insight here.

RReverser:
@paulyoung Babylon doesn't really use ESTree anymore. Although what we're discussing here applies to Flow annotations, which it inherited, too.

kittens:
It's going to be very difficult to add a generic type annotation syntax definitions as Flow and TypeScript diverge in some ways, both syntactically and semantically so they aren't interoperable.

nzakas:
@kittens this spec extension is just limited to the similarities. There is definitely enough commonality to at least take this step.

nzakas:
Can someone please merge this? I'm just documenting what's already implemented with Flow, and I'm trying to write an ESTree-compatible TypeScript parser, so solidifying that this is the way type annotations are done is pretty important.

It's been nearly two months since I opened this, I'd like to get moving. :)

mikesherov:
:+1:

@rreverser merge when you agree please!

RReverser:
@nzakas Thanks!

PR #131, opened 2016-08-03
Async/await has been accepted as a stage 4 proposal.

rictic:
It's scheduled for publishing in ES2017.

See: https://github.com/tc39/proposals/blob/master/finished-proposals.md

### DIFF
--- a/es8.md
+++ b/es8.md
@@ -1,10 +1,10 @@
-# [Async Functions](https://github.com/lukehoban/ecmascript-asyncawait)
+This document specifies the extensions to the core ESTree AST types to support the ES8/ES2017 grammar.
 
 ## Function
 
 ```js
 extend interface Function {
-    async: false;
+    async: boolean;
 }
 ```
 
@@ -13,6 +13,6 @@ extend interface Function {
 ```js
 interface AwaitExpression <: Expression {
     type: "AwaitExpression";
-    argument: Expression | null;
+    argument: Expression;
 }
 ```


es8.md
```diff
@@ -13,6 +13,6 @@ extend interface Function {
 ```js
 interface AwaitExpression <: Expression {
     type: "AwaitExpression";
-    argument: Expression | null;
+    argument: Expression;
```
rictic: I removed `null` here because afaict `await;` isn't legal.

mikesherov:
Thanks for contributing! Seems as if this should move out of experimental. Otherwise, I'll defer to @kittens as to whether the other change is correct.

hzoo:
@mikesherov it got renamed `experimental/async-functions.md → es8.md` so that's good

es8.md
```diff
@@ -1,10 +1,10 @@
-# [Async Functions](https://github.com/lukehoban/ecmascript-asyncawait)
+This document specifies the extensions to the core ESTree AST types to support the ES8/ES2017 grammar.
 
 ## Function
 
 ```js
 extend interface Function {
-    async: false;
+    async: boolean;
```
hzoo: We have the same in babylon https://github.com/babel/babylon/blob/master/ast/spec.md#functions - it makes sense to do since the other booleans are just `boolean` and I guess false was originally there to signify a default

michaelficarra:
These file names are getting ridiculous. All of the files should use the yearly names.

rictic:
Gentle ping! Moving to yearly names is fine by me, but should be separate from this PR IMO.

es8.md
```diff
@@ -1,10 +1,10 @@
-# [Async Functions](https://github.com/lukehoban/ecmascript-asyncawait)
+This document specifies the extensions to the core ESTree AST types to support the ES8/ES2017 grammar.
```
fkling: Is "ES8" really a term that is used? If not and if ES2017 is preferred we might _not_ want to encourage the use of "ES8" by mentioning it here. Same with the file name (although I realize that this is just following existing naming conventions).

Sorry for bikeshedding  🚲  🏠  .

hzoo: @fkling we talk about this in the comments ^ - to use yearly instead

hzoo:
Sure, one of us (I can volunteer) make a pr after to use yearly instead

es8.md
fkling: Oh damn it. There are only 4 comments and I miss the important one 😠

mikesherov:
👍  assuming @hzoo renames afterwards.

mikesherov:
@RReverser merge at will?

PR #134, opened 2016-09-06
Fix UpdateExpression.argument

RReverser:
Couldn't find reasoning for the old one anywhere (backward-compatibility with something?), so I suspect it's just an oversight to include any arbitrary expression.

Thoughts?

### DIFF
--- a/es5.md
+++ b/es5.md
@@ -531,7 +531,7 @@ A unary operator token.
 interface UpdateExpression <: Expression {
     type: "UpdateExpression";
     operator: UpdateOperator;
-    argument: Expression;
+    argument: Pattern;
     prefix: boolean;
 }
 ```


michaelficarra:
We need support for ES5, which means allowing the representation of programs like `a()++;` and `new A = 0;`.

RReverser:
Can you please point me where in ES5 spec `a()++` was allowed?

RReverser:
Even ES3 spec says:

> An  implementation  may  treat  any  instance  of  the  following  kinds  of  runtime  errors  as  a syntax  error  and  therefore report it early: 
...
• Attempts  to  call  PutValue  on  a  value  that  is  not  a  reference  (for  example,  executing  the  assignment  statement

mikesherov:
I believe "an implementation may treat" is the important bit. "May" is different from "Must", and therefore isn't required to be an early error, and therefore may be valid syntax, depending on the implementation.

IMO, we should support parsing this and defer to linters to catch this.

RReverser:
I'm curious how could this get changed in ECMAScript itself between versions. Given that comitee generally accepts only backward-compatible changes (that is, strengthening such rules only when there is no evidence of usage of such syntax in the wild) and made this a strict early error, can we follow it too to avoid inconsistency?

bakkot:
IIRC old versions of Internet Explorer had host functions which returned references, and the committee only made the change once it stopped doing that.

Also, in ES5 (and ES3) it was a runtime error, not an early error, so `if (false) a()++` was a perfectly legal program, and in fact I believe browsers continue to accept it.

--

Incidentally, AssignmentExpressions [also allow arbitrary expressions on the LHS](https://github.com/estree/estree/blob/master/es5.md#assignmentexpression), for the same reason. It would be strange to make this change without likewise changing AssignmentExpressions.

RReverser:
Closing this then, but it's bothering me that we now diverge from the official language specification :/

mikesherov:
@rreverser, how is this in disagreement with the spec? 

If this changed to always be an early error in es6, we could change this in es6.md

RReverser:
@mikesherov Because it was changed as early as ES5 (that's why I started this discussion):

> An implementation **must** treat any instance of the following kinds of errors as an early error: 
...
- Attempts to call `PutValue` on any value for which an early determination can be made that the value is not a `Reference` (for example, executing the assignment statement `3=4`). 

(c) http://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262%205th%20edition%20December%202009.pdf

mikesherov:
> for which an early determination can be made that the value is not a reference.

That seems to be not always possible though, right? And it also seems to not exclude that syntax entirely.

My opinion is this needs to be a parser/interpreter issue, and I don't feel like this brings us out of line with the es5 spec.

How has the text changed in es6 or later?

RReverser:
@mikesherov 

> That seems to be not always possible though, right? 

Yes, as I said, it's vague. As people above pointed, ES5 did allow returning `Reference` from `CallExpression` / `NewExpression`, however, if I understand correctly, `x+2` (`BinaryExpression`) / `!x` (`UnaryExpression`) etc. did fall under "an early determination can be made that the value is not a reference" as there was no way such operations would result in `Reference`, and then implementation **must** prohibit them in LHS.

> How has the text changed in es6 or later?

ES6 is much more specific about it so likely worth separate PR:

> It is an early Reference Error if IsValidSimpleAssignmentTarget of LeftHandSideExpression is false.

mikesherov:
Ok, thanks for explaining. Let's update this in es6

RReverser:
@mikesherov Ok. Last attempt: now that we know the details, perhaps we want to put `Pattern | CallExpression` in ES5 too then? (this would auto-include `new` too)

All the other kinds of expressions in ES5 seem to fall under the "early determination" clause.

mikesherov:
That sounds good to me.

mikesherov:
And maybe some informative text explaining the oddity?

RReverser:
Sure. Let me probably restart that as a separate PR.

PR #135, opened 2016-09-08
Strengthen assignment target definitions.

RReverser:
Reopen of #134 with more precise yet backward-compatible assignment target definitions.

### DIFF
--- a/es5.md
+++ b/es5.md
@@ -531,7 +531,7 @@ A unary operator token.
 interface UpdateExpression <: Expression {
     type: "UpdateExpression";
     operator: UpdateOperator;
-    argument: Expression;
+    argument: Pattern | CallExpression;
     prefix: boolean;
 }
 ```
@@ -584,7 +584,7 @@ A binary operator token.
 interface AssignmentExpression <: Expression {
     type: "AssignmentExpression";
     operator: AssignmentOperator;
-    left: Pattern | Expression;
+    left: Pattern | CallExpression;
     right: Expression;
 }
 ```
--- a/es2015.md
+++ b/es2015.md
@@ -124,7 +124,13 @@ extend interface AssignmentExpression {
 }
 ```
 
-Note that pre-ES6 code was allowed [to pass references around](https://github.com/estree/estree/pull/20#issuecomment-74584758) and so `left` was much more liberal; an implementation might choose to continue using [old definition](https://github.com/estree/estree/blob/master/es5.md#assignmentexpression) if it needs to support such legacy code.
+```js
+extend interface UpdateExpression {
+    argument: Pattern;
+}
+```
+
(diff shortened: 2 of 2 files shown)


mikesherov:
:+1:

es5.md
```diff
@@ -531,7 +531,7 @@ A unary operator token.
 interface UpdateExpression <: Expression {
     type: "UpdateExpression";
     operator: UpdateOperator;
-    argument: Expression;
+    argument: Pattern | CallExpression;
```
michaelficarra: Not just CallExpression. Any expression can be in this position in ES5 and below.

es5.md
```diff
@@ -584,7 +584,7 @@ A binary operator token.
 interface AssignmentExpression <: Expression {
     type: "AssignmentExpression";
     operator: AssignmentOperator;
-    left: Pattern | Expression;
+    left: Pattern | CallExpression;
```
michaelficarra: Same here.

es5.md
RReverser: No, see discussion in previous PRs. In ES5 specifications **must** reject as an early error any LHS that can't evaluate to `Reference` in runtime, and only `CallExpression` and `NewExpression` are such that could result in `Reference` despite being expressions. For literals, unary, binary, conditional and sequence expressions there is no way in spec they can evaluate to `Reference` and thus must be rejected in such position.

michaelficarra: Ah, I didn't know about that early error in ES5. Okay, then I am fine with this as long as we add an item to the "philopsophy" section of the README that states we do not wish to represent programs that would otherwise be syntactically valid if not for their early errors.

bakkot: @michaelficarra that would be a good note to add, but I observe that this is already the case: e.g. in [es2015](https://github.com/estree/estree/blob/master/es2015.md) the LHS of an AssignmentExpression must be a Pattern, but this is only required because of an [early error](https://tc39.github.io/ecma262/#sec-assignment-operators-static-semantics-early-errors).

mikesherov:
@rreverser seems we need to accompany this with a note in philosophy re: early errors. Care to PR that also?

RReverser:
@mikesherov Last time couldn't think of a good wording to specify that, I can give it another try.

adrianheine:
Quick first draft: »ESTree is intended to be able to represent valid ECMAScript scripts and modules. It is not required to support programs that follow the ECMAScript syntactic grammar, but would produce early errors during parsing. See the [introduction to the syntactic grammar in the ECMAScript spec](https://tc39.github.io/ecma262/#sec-syntactic-grammar).«

adrianheine:
We might also want to clarify that it's obviously possible for an invalid source to be representable with ESTree.

nzakas:
@rreverser do you want to move forward with this or should it be closed?

RReverser:
I think it's still useful and valuable, but there was some confusion above as to whether ESTree should represent nodes that wouldn't pass early errors checks.

nzakas:
How would you like to proceed? If you’d like to push forward, I’d suggest summarizing the discussion to this point to make it easier to pick up for others.

PR #137, opened 2016-10-25
Add experimental representation for dynamic import

RReverser:
Adding suggested representation for dynamic `import` expression stage 2 proposal: https://github.com/domenic/proposal-dynamic-import

This one is quite simple and for now will be only under `experimental` folder, but if there are any concerns or suggestions, please let me know :)

cc @domenic @mikesherov @ariya

### DIFF
--- a/experimental/import-expression.md
+++ b/experimental/import-expression.md
@@ -1,0 +1,13 @@
+# [Dynamic Import Expression](https://github.com/domenic/proposal-dynamic-import)
+
+## Import
+
+```js
+interface Import <: Node {
+    type: "Import";
+}
+
+interface CallExpression <: Expression {
+  callee: Expression | Super | Import;
+}
+```
--- a/es5.md
+++ b/es5.md
@@ -667,8 +667,10 @@ A function or method call expression.
 ## NewExpression
 
 ```js
-interface NewExpression <: CallExpression {
+interface NewExpression <: Expression {
     type: "NewExpression";
+    callee: Expression;
+    arguments: [ Expression ];
 }
 ```
 


hzoo:
Just for reference the babylon pr https://github.com/babel/babylon/pull/163 (I linked it in the webpack issue)

Originally was `ImportCallExpression` similar to `ImportExpression` but was changed to be a new `Import` node similar to `Super` https://github.com/estree/estree/blob/master/es2015.md#expressions

mikesherov:
@hzoo any reason y'all chose a new node? The reason Super had a new node was that it took multiple syntatic forms (Call Expression and Member Expression). Not sure what the argument is for creating a new node type when it *seems* like this is allowed anywhere an expression is allowed.

mikesherov:
@hzoo would be great to start here if the goal is eventual interop.

hzoo:
Got it from https://github.com/babel/babylon/pull/163#discussion_r82511862.

> The reason Super had a new node was that it took multiple syntatic forms (Call Expression and Member Expression)

Ah makes sense

cc @loganfsmyth @kesne @ljharb

loganfsmyth:
That's fair, I think at the time of the original discussion on Babylon we hadn't even nailed down if it was restricted to a single argument, so a general CallExpression made sense. Now that it's locked down, we could change that up.

kesne:
I'm happy to make the corresponding changes to Babylon based on what is decided.

ljharb:
(stage 2, not 3, btw)

ljharb:
Not sure how this changes the discussion, but it's highly highly likely that there will be `import.foo` forms in the future - for example, to syntactically provide information about the current module (like `__dirname` does now in CJS).

mikesherov:
@ljharb that does change it alot, because then it *is* more like super. Any public signals about that form you can point to @ljharb ?

ljharb:
https://github.com/rwaldron/tc39-notes/blob/d0c651b358b361b0855cfe7af9ba0b76cab73949/es7/2015-09/sept-24.md has some mentions of "import.context" - I don't think there's concrete plans yet, but I know that any context-sensitive information would need to be reachable through a metaproperty on `import`.

jhnns:
@mikesherov yes, it's similar to `super()`, see https://github.com/domenic/proposal-dynamic-import/issues/8#issuecomment-245938981

RReverser:
Ok, I feel convinced that something like @hzoo suggested (but with `source` field) would be a better way:

```js
interface Import <: Node {
  type: "Import";
  source: Expression;
}
```

Although I don't really like how special-case that we had for `Super` now becomes more common in new proposals instead of sticking with regular expression syntax rules, but that discussion is outside of the ESTree scope anyway.

ljharb:
@RReverser metaproperties (like `super.foo`, `new.target`, `function.sent`, etc) are going to become increasingly common, so there needs to be some pattern to express them.

RReverser:
@ljharb We do have it for metaproperties. I was speaking about speciality of call itself (that keyword is allowed only in callee position and not anywhere else).

mikesherov:
ok @rreverser can you revise PR so we can bikeshed here and then change in appropriate consumers?

ljharb:
`import()` is better than `System.import` precisely because it's syntactic, so it can be a special form that can have special knowledge about usage context.

RReverser:
@mikesherov Updated.

experimental/import-expression.md
```diff
@@ -0,0 +1,13 @@
+# [Dynamic Import Expression](https://github.com/domenic/proposal-dynamic-import)
+
+## ImportExpression
```
hzoo: Should the file name and heading here still be `ImportExpression` or do we also want to call it dynamic import?

mikesherov: good catch.

RReverser:
Btw, yeah - do we want to call it `Import` or `DynamicImport`? First one is shorter (like `Super`) but might be easily confused with import declarations or specifiers.

experimental/import-expression.md
nzakas: It seems like "Import Expression" better follows existing ESTree constructs. "Dynamic" isn't really a syntactic description IMHO.

guybedford:
`Import` does seem a little too generic to me. I think `DynamicImport` would be preferable as clarity should outweigh brevity. Although backwards compatibility will likely be an issue though at this stage.

guybedford:
Another alternative name may be `ContextualImport` to cater to possible metadata properties in future.

ariya:
We ought to agree on the node type:

- `Import`: already in Babel, little too generic, possible confusion with import declaration/specifier
- `DynamicImport`: not really a syntactic description
- `ContextualImport`
- `ImportExpression`: aligned closely with the existing `NewExpression` and `CallExpression`
- `ImportCallExpression`: minor variant of `ImportExpression`
- `ImportCall`: the name in the grammar production in the proposal

mikesherov:
@guybedford, if/when import gets metadata properties, wouldn't `import` in that case by a `metaExpression`? Meaning I think we only need to worry about a bare `import` call here, is that right?

hzoo:
> Sorry if this is unrelated

@ariya FYI, Babel can always change the AST node. We are going to release a Babel 7 where we can change this and in the future for all experimental proposals we can independently version the parser version (more difficult but I believe we have a solution that wouldn't require major bump) and babel plugins/presets. (Babylon will also have an ESTree plugin coming soon so that we are aligned again), although changing all plugins to use it will be another large issue to tackle.

Sounds like `ImportExpression` or `ImportCallExpression` is a 👍 from me for clarity?

ariya:
SGTM. I don't have any strong preference, if I have to pick one it'll probably be `ImportCallExpression`.

RReverser:
`ImportExpression` and `ImportCallExpression` both sound good to me.

RReverser:
Although wait, I thought above we agreed that this is more like `Super` and thus we would use just `Import` to use both in `CallExpression` and `MetaExpression`?

mikesherov:
@rreverser that was my understanding.

mikesherov:
Considering the above input, `DynamicImport` as part of a callexpression or a metaexpression seems to be clearest and most consistent with existing structures.

ariya:
@RReverser @mikesherov I like that. In fact, if we go down that path, there is no for a special new node at all, e.g. `import(something)` can just result in:

```js
{ type: 'CallExpression',
  callee: Identifier { type: 'Identifier', name: 'import' },
  arguments: [ Identifier { type: 'Identifier', name: 'something' } ] }
```

RReverser:
@ariya Yeah but `import` is not really a valid `Identifier`, and has special meaning (allowed in very specific contexts). That's a reason why we introduced separate node for `super` instead of `Identifier` too - it's allowed only in few contexts.

ariya:
@RReverser Oh, are you think of something like this?
```js
{ type: 'CallExpression',
  callee: DynamicImport { type: 'DynamicImport ' },
  arguments: [ Identifier { type: 'Identifier', name: 'something' } ] }
```

RReverser:
@ariya Yeah, just like we do with `Super`.

RReverser:
@domenic This isn't part of the grammar, but AST name. We do want it to be like `super`, but while `Super` name for AST node is quite clear, `Import`, as mentioned above, is little too generic and might be confused with `ImportDeclaration` etc., so would be nice to have more precise name.

domenic:
Doesn't the same problem apply to super? super call vs. super property access.

RReverser:
@domenic No, we use same `Super` node type in both cases.

domenic:
Right, so if you don't find that confusing, I don't see why using `Import` here would be confusing.

ariya:
Sounds like the remaining non-controversial node name (as the `callee`) is `ImportCall`.

RReverser:
@domenic `Import` is confusing not because of context, but rather conflicts with existing `ImportSpecifier`, `ImportDeclaration`, `ImportDefaultSpecifier` and `ImportNamespaceSpecifier` - it might sound like something that is common for all of them, not as a separate syntactic element unrelated to these.

RReverser:
@ariya Except that `ImportCall` would look weird if/when we get metaproperties. Unless you want to name it `ImportCallee` or something.

kesne:
@RReverser If you take `Import` standalone, then sure it could be confusing, but I don't see how it would be confusing in the AST context of being the callee in a `CallExpression`, which is what we're talking about here.

mikesherov:
Yeah, I think we're **safe** to call it `Import`, and I'm not so sure who would be confused by this!

So I'd like then to nail down this proposal: simply one node type `Import`, and differentiate based on whether it's parent node is a CallExpression or a MetaExpression.

ariya:
I'm fine with `Import` as the node name.

RReverser:
Sounds great. I'll update PR when back on Monday.

mikesherov:
Ok, great! Good news @hzoo, seems like we'll converge on what Babylon already has!

ariya:
Slightly relevant: can `Import` be the callee of `NewExpression` as well and not only `CallExpression`?

The [proposal](https://tc39.github.io/proposal-dynamic-import/#sec-left-hand-side-expressions) only lists _ImportCall_ as an addition to the production grammar of _CallExpression_.

Kovensky:
It wouldn't make much sense for `Import` to be the callee. _if_ it were allowed, it'd either be the same footgun as `new require('foo')` (calls `require` as a constructor) or it'd construct a module record, which is not callable or constructible.

guybedford:
The main confusion I think that can happen with `Import` is that for a new developer looking at a transform using a visitor pattern, it may not be obvious that `Import` is visiting a dynamic import, and not a static import. But this confusion isn't really a concern of this spec I guess, and `Import` does seem to make complete semantic sense.

RReverser:
Updated and removed relation between `CallExpression` and `NewExpression` - Indeed, with introduction of special callees it doesn't make sense any more.

RReverser:
@guybedford Developer needs either to look at AST definition or use TypeScript definitions anyway. In that case, it will be crystal clear that `Import` accepts `Expression` while regular `ImportDeclaration` etc. only accept static strings.

mikesherov:
Ok, thanks everyone for your input. I thumbed the PR. Feel free to land @rreverser

PR #138, opened 2016-11-08
Create async-iteration.md

hzoo:
https://github.com/tc39/proposal-async-iteration (stage-3)

Original babylon PR: https://github.com/babel/babylon/pull/17

`ForAwaitStatement`

Other suggestions were:
- `ForOfStatement` with `"await": true`
- name as `ForAwaitOfStatement`

### DIFF
--- a/experimental/async-iteration.md
+++ b/experimental/async-iteration.md
@@ -1,0 +1,9 @@
+# [Async Iteration](https://github.com/tc39/proposal-async-iteration)
+
+## ForAwaitStatement
+
+```js
+interface ForAwaitStatement <: ForInStatement {
+  type: "ForAwaitStatement";
+}
+```
--- a/experimental/async-iteration.md
+++ b/experimental/async-iteration.md
@@ -1,9 +1,9 @@
 # [Async Iteration](https://github.com/tc39/proposal-async-iteration)
 
-## ForAwaitStatement
+## ForOfStatement
 
 ```js
-interface ForAwaitStatement <: ForInStatement {
-  type: "ForAwaitStatement";
+extend interface ForOfStatement {
+  await: boolean;
 }
 ```


nzakas:
I think `ForAwaitOfStatement` for closely mirrors what we do with other loop statements.

RReverser:
> ForOfStatement with "await": true

I like this one a bit more - it's not as big difference as between `for ... in` vs `for ... of`, so can totally be expressed via a flag.

mikesherov:
I like flags here as well. I can't think of many rules that wouldn't apply to the await version equally. Not producing a new node type allows here easier transformation.

nzakas:
Using the same node type means a breaking change for consumer who are expected `for-of` to not contain `await` after `for`. For tools like ESLint, that could have rippling effects through rules (both core and custom).

If I'm in the minority here, so be it, but my position is still that this should we a new node type for the easiest ecosystem integration.

RReverser:
@nzakas That's exactly the same what we did for `FunctionExpression` / `FunctionStatement` though. It's an extension, not breaking representation of existing syntax. Moreover, most visitors that apply to `ForOfStatement`, will apply to async version of it too, so I don't see this as a big problem, rather a benefit.

mikesherov:
@nzakas we experienced the same pros and cons with `shorthand`, `method`, and `computed` properties: wide Syntactic differences with same semantic meaning.

Yes, there is a risk but most rules will just work.

Thoughts?

nzakas:
> we experienced the same pros and cons with shorthand, method, and computed properties

Yes, these have caused me a ton of pain since ES6 was finalized. I regard these as the worst part of ESTree and was hoping we could carve out a better path for new syntax so as to avoid further pain. 

But I also don't have the energy to debate, so as I said, if I'm the only one concerned, please disregard.

mikesherov:
@hzoo what are your thoughts on breaking Bablyon to using `ForOfStatement` with `await: true`? What were the arguments against it? Same as @nzakas?

hzoo:
We can do the breaking change if everyone thinks it's better since we are doing Babel 7 now

> I've chosen to output a new node type: "ForAwaitStatement". Another approach would be to output a "ForOfStatement" with an "await" property set to true.

cc @zenparsing

danez:
I don't have a strong opinion for separateNode vs flag, and from a babel standpoint it doesn't matter as much, as switching to a flag is just one more check in the visitor `node.await === true`.

Also agree with @hzoo, we could do this breaking change with 7.0.

mikesherov:
Ok, so it seems as if property on existing node is preferred by @rreverser and myself, with @nzakas dissenting but ok if we proceed, and @hzoo being ok with this change.

@ariya, any opinion here? If there's no other discussion to be had, I'd like to move forward on this in the next few days.

ariya:
I don't have any strong preference one way or another.

mikesherov:
Ok. @hzoo let's amend this PR as described and plan for breaking Babylon change. Thanks for accommodating!

hzoo:
Sorry, made the fixes!

and Babel PRs:
https://github.com/babel/babylon/pull/349
https://github.com/babel/babel/pull/5321

hzoo:
I'l merge this if it's ok, we already made the changes for Babel 7

RReverser:
Oach, I missed the "feel free to land" bit.

mikesherov:
Thanks!

PR #152, opened 2017-02-12
Add directive

### LINKED ISSUE #6: Extend ExpressionStatement to indicate a directive prologue

danez:
This fixes #6 and is a replacement for #99 

I changed the text to exclude quotes in the directive, fixed some formatting and change the `FunctionBody` node to not have a type as it is inherited anyway.

### DIFF
--- a/es5.md
+++ b/es5.md
@@ -137,7 +137,7 @@ these flags `value` will be `null` as the regex can't be represented natively.
 ```js
 interface Program <: Node {
     type: "Program";
-    body: [ Statement ];
+    body: [ Directive | Statement ];
 }
 ```
 
@@ -149,7 +149,7 @@ A complete program source tree.
 interface Function <: Node {
     id: Identifier | null;
     params: [ Pattern ];
-    body: BlockStatement;
+    body: FunctionBody;
 }
 ```
 
@@ -174,6 +174,17 @@ interface ExpressionStatement <: Statement {
 
 An expression statement, i.e., a statement consisting of a single expression.
 
+## Directive
+
+```js
+interface Directive <: ExpressionStatement {
+    directive: string;
+}
+```
+
+A directive from the directive prologue of a script or function.
+The `directive` property is the raw string source of the directive without quotes.
+
 ## BlockStatement
 
 ```js
@@ -184,6 +195,16 @@ interface BlockStatement <: Statement {
 ```
 
 A block statement, i.e., a sequence of statements surrounded by braces.
+
+## FunctionBody
+
+```js
+interface FunctionBody <: BlockStatement {
+    body: [ Directive | Statement ];
+}
+```
+
+The body of a function, which is a block statement that may begin with directives.
 
 ## EmptyStatement
 


mikesherov:
As discussed in #6...

The benefits here are:
1. Already exists in Esprima, Babylon, and Flow.
2. Directive checks (simplified) become: 
```
const isStrict = body.length && body[0].directive == 'use strict';
// instead of
const isStrict = body.length && body[0].expression && body[0].expression && body[0].expression.value == 'use strict'; `.

Concerns are:
1. Mutations on function body may accidentally cause a directive to be in a non-directive spot, causing the tree to be out of sync. This is partially mitigated by the fact that a mutation that does this is likely already doing the wrong thing.
2. May run a foul of the "Duplicative" design principle. This is possibly Mitigated by the fact that duplicity refers to duplicity on the same node.

Looking to get thumbs ups from @rreverser and @hzoo here.

**Comment by michaelficarra** [comment]
@mikesherov The directive check is not that simple. You have to look through the whole prologue.

An unlisted benefit is that you have access to the raw value so you can, for the first time ever, actually implement a correct directive test on an estree AST.

**Comment by mikesherov** [comment]
@michaelficarra just trying to illustrate a simplified version of the difference. Wasn't trying to be 100% accurate.

Anyway, the raw point is a good one too.

**Comment by zubuzon** [comment]
Is this settled in stone now, so it's safe to implement it, or will it be further modifications? I noticed the discussion has been going on for more than 2 years.

**Comment by mikesherov** [comment]
@zubuzon, thanks for asking. It's essentially settled. Needs feedback from @rreverser though.

**Comment by RReverser** [comment]
@mikesherov Well, you know my concerns and that this somewhat violates our declared philosophy. Other than that, we'll adapt if this is accepted.

**Comment by mikesherov** [comment]
Right, but as discussed, there isn't really a good alternative that is BC and removes the duplicative concern. We could also just say forget it. Neither Esprima nor Acorn have adopted this yet, right?

**Comment by RReverser** [comment]
@mikesherov 

> Neither Esprima

I thought above you said

> Already exists in Esprima, Babylon, and Flow.

**Comment by mikesherov** [comment]
Ugh, I'm getting confused. Ignore me. I'll find out for sure and get my facts straight before adding more noise here.

**Comment by zubuzon** [comment]
I see. Still a lot of confusion. BTW. I noticed for Esprima - which also maintains the raw property on the AST node - there will be duplications. Both raw and the directive node contains the same thing.  Was that what you was thinking about, @mikesherov?

**Comment by danez** [comment]
To add more infos here, this is the adoption rate for this flag:

| name | y/n |
| ------------- | ------------- |
| acorn | no |
| babylon (with estree plugin) | yes |
| espree | no |
| esprima | yes |
| flow | yes |

**Comment by danez** [comment]
Can we make progress on this?
In my opinion this is a good solution and I'm also thinking about proposing to change babylon back to this behaviour.

A correct check for "use strict" with this field is : (which is not super complicated)

```js
function isStrictBody(node) {
  for (const statement of node.body.body) {
    if (statement.directive == null) {
     break;
    } else if (statement.directive === "use strict") {
      return true;
    }
  }

  return false;
}
```

//cc @hzoo @RReverser @mikesherov @michaelficarra

mikesherov:
@RReverser let's call this accepted considering everyone else already implements it aside from Acorn and this is unlikely to ever change or have a better solution.

RReverser:
*sigh* okay...

mikesherov:
Thanks @danez for the patience and pushing this forward. cc @nzakas re: any changes that may be required in espree, although I think there might be none?

PR #157, opened 2017-03-15
Add undefined as a possible Literal value

zzzev:
`undefined` should be a type of Literal. Currently, there's no mention of `undefined` anywhere in the ESTree spec.

As a result, parsers like Esprima treat `undefined` as an identifier with the name "undefined," because that's the only legal way to express it given the current ESTree type definitions. You can see this behavior in the Esprima demo page here: http://esprima.org/demo/parse.html

I suspect there may be some additional reasoning on why this is the way it is currently (perhaps undefined is not a legal value in this syntax?), but I've been unable to find any references to `undefined` in this repo or associated issues/pull requests, or in Esprima's repo/associated stuff.

### DIFF
--- a/es5.md
+++ b/es5.md
@@ -111,7 +111,7 @@ An identifier. Note that an identifier may be an expression or a destructuring p
 ```js
 interface Literal <: Expression {
     type: "Literal";
-    value: string | boolean | null | number | RegExp;
+    value: string | boolean | null | number | RegExp | undefined;
 }
 ```
 


zzzev:
After [reading a bit more](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) on this subject, it looks like my assumption was actually incorrect and this behavior is correct for backwards compatibility with older browsers.

Sorry for the bad PR; would love to hear any additional thoughts the team has on ways to more clearly distinguish undefined since it does not behave like a "normal" identifier in modern browsers.

loganfsmyth:
`var undefined = 4` is valid, even in modern browsers. There is no way of knowing if it is really an undefined value without analyzing the full scope of the program. The only thing special about `undefined` is that there is an immutable global variable called `undefined` so when you use it in your day to day code, you're getting the value of that global variable.

gibson042:
https://jsfiddle.net/befkbxww/

PR #166, opened 2017-05-19
Template literal revision

isiahmeadows:
The [template literal revision](https://tc39.github.io/proposal-template-literal-revision/) recently accepted for ES2018 has no ESTree support. It currently requires that nodes that would otherwise be invalid (e.g. <code>latex \`\newcommand{\unicode}{\textbf{Unicode!}}\`</code>) will have their non-raw component evaluate to `undefined` when they don't exist, but ESTree needs updated to support the possibly non-existent value.

### DIFF
--- a/es2018.md
+++ b/es2018.md
@@ -1,0 +1,11 @@
+This document specifies the extensions to the core ESTree AST types to support the ES2018 grammar.
+
+```js
+extend interface TemplateElement {
+    value: {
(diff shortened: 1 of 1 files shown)


isiahmeadows:
Closing in favor of [this](https://github.com/estree/estree/blob/master/experimental/invalid-tagged-template-escapes.md) being moved, since I didn't see it.

PR #172, opened 2017-10-05
Remove redundant Directive | Statement instances

robsimmons:
Because a Directive is an ExpressionStatement and an ExpressionStatement is a Statement, the use of Directive | Statement option was redundant. This change reverts back to a simpler description that is equivalent to the current one.

This is an alternative to PR #173

### DIFF
--- a/es5.md
+++ b/es5.md
@@ -137,7 +137,7 @@ these flags `value` will be `null` as the regex can't be represented natively.
 ```js
 interface Program <: Node {
     type: "Program";
-    body: [ Directive | Statement ];
+    body: [ Statement ];
 }
 ```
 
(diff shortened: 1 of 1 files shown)


RReverser:
Alternative could be to make `Directive` inherit from `Node`, so that it wouldn't be allowed inside of loops etc., but instead would be explicitly allowed via `Directive | Statement` only at the top level of `Program` and `Function`.

RReverser:
Just noticed https://github.com/estree/estree/pull/173/files which seems to do exactly that.

michaelficarra:
I'm okay with just rejecting this PR in favour of #173.

PR #173, opened 2017-10-05
Make Directive | Statement comments meaningful

robsimmons:
Because a Directive is an ExpressionStatement and an ExpressionStatement is a Statement, the use of Directive | Statement option was redundant. This change attempts to capture the intent of the Directive | Statement, by making it clear that while Directive is structurally a subtype of ExpressionStatement, the directive type is not a valid addition to any ExpressionStatement, just ones that are children of a Function or Program.

This is an alternative to PR #172.

### DIFF
--- a/es5.md
+++ b/es5.md
@@ -179,7 +179,7 @@ An expression statement, i.e., a statement consisting of a single expression.
 ```js
 interface Directive <: Node {
     type: "ExpressionStatement";
-    expression: Expression;
+    expression: Literal;
     directive: string;
 }
 ```
--- a/es5.md
+++ b/es5.md
@@ -177,7 +177,9 @@ An expression statement, i.e., a statement consisting of a single expression.
 ## Directive
 
 ```js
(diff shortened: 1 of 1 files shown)


es5.md
```diff
@@ -177,7 +177,9 @@ An expression statement, i.e., a statement consisting of a single expression.
 ## Directive
 
 ```js
-interface Directive <: ExpressionStatement {
+interface Directive <: Node {
+    type: "ExpressionStatement";
+    expression: Expression;
```
RReverser: I think it's reasonable to have `Literal` here.

robsimmons: Agreed

mikesherov:
Thanks @robsimmons

ariya:
👍 , just like [in Esprima](https://github.com/jquery/esprima/blob/master/src/nodes.ts#L289).

RReverser:
@ariya Any reason Esprima doesn't limit `Directive::expression` to `Literal` like we did here?

PR #174, opened 2017-11-13
Add OptId{Class,Function}Declaration for ExportDefaultDeclaration

### LINKED ISSUE #98: Allow null for FunctionDeclaration and ClassDeclaration id

adrianheine:
Closes #98. This implementation was initially suggested by Sebastian McKenzie.

### DIFF
--- a/es2015.md
+++ b/es2015.md
@@ -395,9 +395,19 @@ An exported variable binding, e.g., `{foo}` in `export {foo}` or `{bar as foo}` 
 ### ExportDefaultDeclaration
 
 ```js
+interface AnonymousDefaultExportedFunctionDeclaration <: Function {
+    type: "FunctionDeclaration";
+    id: null;
+}
+
+interface AnonymousDefaultExportedClassDeclaration <: Class {
+    type: "ClassDeclaration";
+    id: null;
+}
+
 interface ExportDefaultDeclaration <: ModuleDeclaration {
     type: "ExportDefaultDeclaration";
-    declaration: Declaration | Expression;
+    declaration: AnonymousDefaultExportedFunctionDeclaration | FunctionDeclaration | AnonymousDefaultExportedClassDeclaration | ClassDeclaration | Expression;
 }
 ```
 


es2015.md
```diff
@@ -397,7 +412,7 @@ An exported variable binding, e.g., `{foo}` in `export {foo}` or `{bar as foo}`
 ```js
 interface ExportDefaultDeclaration <: ModuleDeclaration {
     type: "ExportDefaultDeclaration";
-    declaration: Declaration | Expression;
+    declaration: OptIdFunctionDeclaration | OptIdClassDeclaration | VariableDeclaration | Expression;
```
mysticatea: I think `VariableDeclaration` is disallowed here: https://www.ecma-international.org/ecma-262/8.0/#prod-ExportDeclaration

adrianheine: Agreed.

michaelficarra:
I don't know what it means in our pseudo-IDL for an interface to extend another but have the same property name. In all other cases, an interface extending another means instances of the interface are necessarily instances of the extended interface. That wouldn't be so in this case. :-1:

michaelficarra:
FWIW, what I think you're trying to do can be done by copy/pasting the extended interfaces and adding the `| null` to the `id` property, but then would that node even be useful? If it can be used in exactly one place, it should just be subsumed by that consumer.

adrianheine:
@michaelficarra I understand your concern. What we could do is turn it around and make `id` nullable in the base interface, but have another interface with non-nullable `id` and use that everywhere except for default exports. That would still mean that an extending interface re-defines a property of a parent interface, but at least the child interface fulfills the parent interface's contract.

How would »subsuming by that consumer« look like?

michaelficarra:
@adrianheine Yeah, that inversion should work.

By "subsumed", I was trying to describe something like this:

```
interface A {
  propB: B;
  propC: C;
}

// this interface is only ever used in A
interface B {
  x: X;
}

// this interface is used in more than one place
interface C {
  y: Y;
}

// so A should subsume B (in this scenario, B has been deleted)
interface A {
  x: X;
  propC: C;
}
```

In concrete terms, I was suggesting the `ExportDefaultDeclaration` subsume the unnamed declaration nodes. Since there's more than one, we would have many `export default`-related AST nodes instead.

adrianheine:
I don't feel like I have enough experience with this project to contribute something meaningful here (but I'll try anyway because I like specs that match reality).

acorn and probably all the other parsers nowadays emit for `export default function () {};` a `ExportDefaultDeclaration` with a `FunctionDeclaration` as `declaration`. In comparison to `Function`, `FunctionDeclaration` is a `Declaration` (which doesn't matter for this use-case), has a non-nullable `id` (which is wrong in this use-case) and `"FunctionDeclaration"` as `type`.

Instead, `ExportDefaultDeclaration` could also emit a, let's say `DefaultExportedFunctionDeclaration` with

```
interface DefaultExportedFunctionDeclaration <: Function {
    type: "FunctionDeclaration";
}
```

@kittens, @mikesherov, @hzoo, what do you say?

adrianheine:
```
interface DefaultExportedFunctionDeclaration <: Function {
    type: "FunctionDeclaration";
    id: null;
}

```

michaelficarra:
@adrianheine Perfect. Thanks for sticking with this.

adrianheine:
I updated the commit to what we discussed.

Issue #169, opened 2017-09-13
BigInt


jamiebuilds:
Babylon [represents](https://github.com/babel/babylon/pull/588) BigInt as:

```js
{
  type: "BigIntLiteral",
  extra: {
    rawValue: "0b101011101",
    raw: "0b101011101n"
  },
  value: "0b101011101"
}
```

But it's not obvious how estree will represent them.

- Right now `Literal` is overloaded to mean different things depending on the type of `node.value`
- Since strings are already taken by strings, there will need to be something to distinguish strings and big ints

Options:
1. New property on `Literal` more accurately describing which type it is. i.e
```js
{ 
  type: "Literal",
  kind: "string" | "number" | ... | "bigint"
  value: string | number | boolean | ...
}
```
2. Extend the `Literal` type with a new `BigIntLiteral` type

I figure people will be most in favor of 2 since there's already a RegExpLiteral node.

michaelficarra:
I'm for option 2 to be consistent with RegExp literals.

```js
{ type: "Literal", value: 0n, bigint: "0n" }
```

mikesherov:
I am also for number 2, even if it would be "consistent" with the majority of ESTree literals (let's not repeat mistakes of the past). It would be a bit surprising for consumers of ESTree, but this is a cleaner future.

@RReverser @ariya any thoughts?

ariya:
👍  for `bigint` property in the `Literal` node.

adrianheine:
In #179 the question of how to handle the `value` property on host engines that don't support bigint arised.

Issue #64, opened 2015-03-15
ASTs should be JSON compatible


gibson042:
`Literal#value` appears to be the only potentially incompatible property, and even there string/boolean/null are naïvely covered. Regular expressions are not valid JSON, but that's already been dealt with as well. So we're left with numbers, which fortunately are specified as double-precision 64-bit IEEE 754 values and fit comfortably within JSON's infinite precision—the singular exception being _numerically-expressed_ values greater than approximately 2^1024, which are infinite in Javascript.

Embracing 27d0e173d7b220044c5fa05da1e9c3cfcf96a41b would probably mean introducing `NumberLiteral <: Literal` with properties like `number: { infinite: true }` or `number: { from: ".2e309" }`/`number: { precise: "0x<256+ 'F's>" }`/etc. (Non-exhaustive) alternatives include:
* formalizing `raw`
* eliminating the wrapper object (e.g., `infiniteNumber: boolean` instead of `number: { infinite: boolean }`)
* closing this ticket to let each toolchain make independent decisions about this edge case

At any rate, though, the topic has been raised frequently enough that I considered it worth documenting.

sebmck:
:+1: formalizing `raw` (#14)

impinball:
I do believe `raw` is the solution to this one, since I personally have found a ton of uses for it, including testing for hex numbers and specifying what escodegen needs to output in a number.

michaelficarra:
I don't think raw is best here. It adds a lot of burden for something that can be fixed much more simply with one of the above suggestions.

RReverser:
@michaelficarra That would be true if this would be the only reason for incorporating `raw`, but looks like it will be formalized in any case, so we can reuse it for this issue as well instead of duplicating info through some other property.

adrianheine:
How about dropping the `value` field?

1. With bigint and extensions to regular expressions, it becomes more uncertain whether a JS host can represent a given JS literal. For hosts implemented in other languages this is even more true. We won't be able to represent them in JSON.
2. Creating the value takes time and uses space while I expect it to be used rarely (at least by acorn users).

marijnh:
Seconding the suggestion to drop `value`. It awkwardly ties the format to the host platform, making it difficult to create ASTs on anything but JavaScript platforms that already support the (regexp, bignum) features that are being parsed.

PR #179, opened 2017-12-19
Add BigInt

adrianheine:
This is my try at representing what I understood to be the agreed-upon solution in #169.

### DIFF
--- a/experimental/bigint.md
+++ b/experimental/bigint.md
@@ -1,0 +1,12 @@
+# [BigInt](https://github.com/tc39/proposal-bigint)
+
+```js
+extend interface Literal {
+  value: string | boolean | null | number | RegExp | bigint;
+}
+
+interface BigIntLiteral <: Literal {
+  value: bigint | null;
+  bigint: string;
+}
+```


experimental/bigint.md
```diff
@@ -0,0 +1,7 @@
+# [BigInt](https://github.com/tc39/proposal-bigint)
+
+```js
+interface BigIntLiteral <: Literal {
+  bigint: string;
```
michaelficarra: Other `Literal`s use `value` as the property name.

RReverser: As far as I understand, the idea here is to have it similar to `RegExpLiteral` where `value` might or might not be representable on the target runtime so another field is needed to represent the parsing result in these cases.

RReverser: Actually, you were the one who suggested that in https://github.com/estree/estree/issues/169#issuecomment-329059842

michaelficarra: Ah, that's right, there's no `"type"` or anything distinguishing it from other `Literal`s. Got it. Okay, what should the `value` of these nodes be? Currently bigints are not allowed in the `value` field.

adrianheine: You suggested to use a BigInt literal for `value`. For that we would need to amend the `Literal` interface (`extend interface Literal { type: string | boolean | null | number | RegExp | bigint;`) and add `value: bigint` to this interface, right?

michaelficarra: Yep, that should do it.

adrianheine: I pushed it, but with property `value`, not `type`.

experimental/bigint.md
```diff
@@ -0,0 +1,12 @@
+# [BigInt](https://github.com/tc39/proposal-bigint)
+
+```js
+extend interface Literal {
+  value: string | boolean | null | number | RegExp | bigint;
+}
+
+interface BigIntLiteral <: Literal {
+  value: bigint;
+  bigint: string;
```
j-f1: How about `raw` and `cooked`? (i.e.`raw` would be `"12345n"` and cooked would be `12345n` (as a `BigInt`)?

adrianheine: `cooked` only exists with template literals. `raw` is not part of the spec for (non-template) literals (see #14).

As for platforms without BigInt support, I've been asking this myself, too. Babylon doesn't emit the value, so I'm not aware of any precedents. I can think of a) emitting null b) casting to a regular number, which is lossy.

j-f1: FYI “platforms without `BigInt` support” includes (and will include, for all eternity<sup>[*citation needed*]</sup>) JSON.

adrianheine: You are right, just as regular expressions. There's a discussion in #64 about JSON.

RReverser: I think emitting `null` if bigint is not in safe integer space makes sense and would be in line with regular expression literals.

michaelficarra: Why safe integer space? Use a bigint if the platform supports it (regardless of magnitude), `null` otherwise.

adrianheine: I've made `value: bigint | null` for now, but I would still prefer to drop or phase out `value` as explained in https://github.com/estree/estree/issues/64#issuecomment-357434955.

goodmind:
Should it be null if integer isn't in safe space or always null? (for JSON dump of AST)

Issue #177, opened 2017-12-17
Template Literals Revision is stage4


adrianheine:
I suppose that means we should agree on `experimental/invalid-tagged-template-escapes.md` and put it into a new `es2018.md`?

PR #186, opened 2018-01-28
Move invalid-tagged-template-escapes to es2018

mysticatea:
> https://github.com/tc39/proposals/commit/cba6aa3c8ace8d87c6fb8a0e9b9262240815ce34#diff-79a53ca5c00fe0d90f18cd5237280da3

Lifting Template Literal Restriction has arrived at Stage 4.
This PR moves it to es2018.md from the experimental directory.


adrianheine:
Closes #177.

mysticatea:
@adrianheine Thank you. I added `Closes` sign in the post.

es2018.md
```diff
@@ -10,6 +10,19 @@ extend interface ObjectExpression {
...
+```js
+extend interface TemplateElement {
+    value: {
+        cooked: string | null;
+        raw: string;
+    };
+}
+```
+
+Lifting Template Literal Restriction, e.g., ``tag`\unicode and \u{55}` ``
```
j-f1: How about:

> If the template literal is tagged and the text has an invalid escape, `cooked` will be null, e.g., ``tag`\unicode and \u{55}` ``

mysticatea: Sounds good! I updated it.

es2018.md
```diff
@@ -10,6 +10,19 @@ extend interface ObjectExpression {
...
+```js
+extend interface TemplateElement {
+    value: {
+        cooked: string | null;
+        raw: string;
+    };
+}
+```
+
+If the template literal is tagged and the text has an invalid escape, `cooked` will be `null`, e.g., ``tag`\unicode and \u{55}` ``
```
michaelficarra: `\u{55}` is a bad example since it was already valid. Prefer invalid hex.

PR #189, opened 2018-05-22
move experimental/optional-catch-binding.md to es2019.md

mysticatea:
This PR moves [Optional catch binding](https://github.com/tc39/proposal-optional-catch-binding) syntax to es2019.md from the experimental directory because it arrived at stage 4: https://github.com/tc39/proposals/commit/83f2325bbf9804052cf371d3fdcd2a9b420715c4


mysticatea:
Thank you for the review. I fixed to use `extend`

PR #198, opened 2019-06-04
add Dynamic Imports and BigInt as ES2020

### LINKED ISSUE #169: BigInt

### LINKED ISSUE #197: Revisit Dynamic Imports
Hello.

Currently, AST for dynamic imports is using `CallExpression`: https://github.com/estree/estree/blob/2c161ae714c1b23198be3be08c930e55a621c187/experimental/import-expression.md
But I think an independent node type is better:

```ts
interface ImportExpression <: Expression {
    type: "ImportExpression";
    source: Expression;
}
```

For some reasons:

- `source` property is consistent with [ImportDeclaration](https://github.com/estree/estree/blob/2c161ae714c1b23198be3be08c930e55a621c187/es2015.md#importdeclaration) to express importing source.
  On the other hand, it's hard to read that `arguments[0]` express the importing source.
- Searching `CallExpression`s which have `Import` node in that `callee` is not convenient to analyze dependencies.
- The `arguments` property doesn't fit Dynamic Imports syntax.
    - The syntax cannot have multiple arguments. (i.e. `import()` and `import(a, b)` are syntax error)
    - The syntax cannot have trailing commas. (i.e. `import(source,)` is syntax error)
- [EDIT] `Import` node is not consistent with `import.meta` meta property. Existing `MetaProperty` node represents `import.meta` without `Import` node. The `Import` node will confuse us in the future.

Thoughts?

Thank you.

----

- Previous discussion: https://github.com/estree/estree/pull/137

mysticatea:
> - https://github.com/tc39/proposals/commit/b04295ff4af51c143c62df62947ae7cb25759691
> - https://github.com/tc39/proposals/pull/217 (the [meeting notes](https://github.com/tc39/tc39-notes/blob/master/meetings/2019-06/june-4.md#bigint-to-stage-4) have the conclusion "Stage 4 acceptance")

Dynamic Imports and BigInt syntaxes have arrived at Stage 4. This PR adds those syntaxes as ES2020.

> [es2020.md](https://github.com/estree/estree/blob/a7ed03d1a59982e956823fb084072ba1848fa137/es2020.md)

- I extended `Literal` as separated from `BigIntLiteral` because of consistent. [Literal](https://github.com/estree/estree/blob/master/es5.md#literal) has `RegExp` value and `RegExpLiteral` extends only `regex` property.
- [EDIT] ~~I represented Dynamic Imports as `CallExpression` as same as the current experimental spec. However, I think that another way such as `ImportExpression` is better: https://github.com/estree/estree/issues/197. Thoughts?~~ I changed the AST of Dynamic Imports because of #197. I believe that the `Import` node way will confuse us in the future.

---

Fixes #197, fixes #169, closes #179.


adrianheine:
There's already #179 for BigInts (but I don't care which one is merged). I'm still in favor of not having `value` for them in the spec.

mysticatea:
Because that was adding the spec into experimental directory, I have written this to move both syntaxes to es2020.md.

I prefer having `value`. The `Literal` node has `value` property, so the derived interface must have `value` property. We have experienced this behavior already when new syntaxes were added in RegExp. Indeed, nullable is odd, but I'm not worried because it's temporary.

michaelficarra:
@mysticatea In a bunch of places, you've mistyped `BigInt` as `BitInt`.

es2020.md
```diff
@@ -0,0 +1,48 @@
...
+## BigIntLiteral
+
+```js
+interface BigIntLiteral <: Literal {
+  bigint: string;
+}
+```
+
+- `bigint` property is the string representation of the `BitInt` value.
+  It includes the suffix `n`.
```
michaelficarra: Why?

mysticatea: No reason :D

Just I thought the spec should clear which the value of `bigint` property is.
I don't mind that it includes or doesn't include `n`.

mysticatea:
I'm leaning toward, we should change the representation of dynamic imports: #197.

es2020.md
michaelficarra: I think it's great that the spec is explicit about it. I would prefer to omit the `n`. You just need to be able to reconstruct the value from this field, and there's no point in adding an `n` to the end. Besides, BigInt's `toString` also omits the `n`.

mysticatea: Sounds reasonable. I will change this sentence.

mysticatea:
What should I do to advance this PR?

mysticatea:
@RReverser @ariya would you take a look this PR?

es2020.md
```diff
@@ -0,0 +1,43 @@
...
+  type: "ImportExpression";
+  source: Expression;
+}
+```
+
+- `ImportExpression` node represents Dynamic Imports such as `import(source)`.
+  The `source` property is the importing source as similar to [ImprotDeclaration]
+  node, but it can be an arbitrary expression node.
+
+[ImprotDeclaration]: es2015.md#importdeclaration
```
michaelficarra: Typo: ImprotDeclaration

es2020.md
```diff
@@ -0,0 +1,43 @@
...
+
+```js
+interface BigIntLiteral <: Literal {
+  bigint: string;
+}
+```
+
+- `bigint` property is the string representation of the `BigInt` value.
+  It doesn't include the suffix `n`.
+- In environments that don't support `BigInt` values, `value` property will be
```
michaelficarra: Instead of saying this in prose, I would add a `value: null` entry to the interface description.

mysticatea: The type of `value` property is `string | boolean | null | number | RegExp | bigint`, so it's nullable already.

es2020.md
mysticatea: Oops, thank you.

es2020.md
mysticatea: The purpose of this sentence is to describe we cannot use `node.value === null` to distinguish if a node is `null` literal or not.

loganfsmyth: This seems like it could get pretty confusing if there are two possible representations of the field. Wouldn't it be much nicer to do
```
interface BigIntLiteral <: Literal { }
```
with the expectation that `.value` is the string representation when `.type === "BigIntLiteral"`?

mysticatea: I agree that this is inconvenient. But we did the same thing in the past: es2015 and es2018 RegExp enhancements. So this is following precedents.

mysticatea:
What should I do to advance this PR?

mercmobily:
Are ther any more steps needed/revisions before this PR can be merged? Adding this in case this slipped

mercmobily:
I am typing here as ESlint depends on this to get dynamic imports happening! 

https://github.com/eslint/eslint/issues/11803

aladdin-add:
any news?

RReverser:
@adrianheine Still would like your input on your PR vs this one.

mikesherov:
I don't think we can land this as is. Unless I'm mistaken, both Webpack and Acorn have adopted the current experimental design for dynamic import. Seems it's too late and not worth the effort to undo the change. Just my 2 cents here.

mysticatea:
I know Acorn's release was a mistake. I think they will make a major release after this PR merged.

mikesherov:
@mysticatea and what about Webpack? This would need to get merged and released before Webpack 5 was released.

mysticatea:
The AST was experimental. I believe we should fix it before getting stable, rather than leaving the pain for the future.

> This would need to get merged and released before Webpack 5 was released.

I hope this PR to be merged as soon as possible. I have been waiting for the decide of ESTree team since a month ago. I'm just waiting.

mikesherov:
ok, so:

Approvals needed:
[x] @mysticatea I suppose you serve as approval from the ESLint side of things.
[x] @RReverser speaks for Acorn. 
[ ] @adrianheine still waiting on your approval as per @RReverser's comment.
[ ] @michaelficarra you've reviewed. Would be nice to get an explicit approval from you.

People to inform:
* Pinging @wsokra as a heads up that there will be an AST change coming up to hopefully get it into Webpack 5.
* And pinging @hzoo on the Babel side of things that this change is coming.

aladdin-add:
friendly ping @adrianheine

mysticatea:
We have over a month (40 days) since the last commit. Is there any chance to be merged?

mikesherov:
@RReverser I think we should merge this one and also credit @adrianheine in the commit message. The two pull requests are fundamentally the same thing (with respect to BigInt), no? Alternatively, we could land the dynamic import stuff (since we have wide agreement here), and slice off BigInt until Adrian replies.

Either way it's time to move this forward :-)

mikesherov:
@RReverser once you're settled in at your new gig, perhaps you can merge this PR, or are we still waiting on anyone?

mysticatea:
We have over two months since the last commit. Is there any chance to be merged?

mikesherov:
Thanks @RReverser! @mysticatea do have any PRs open against Acorn acknowledging this change?

mysticatea:
Thank you for merged!

Yes, I have https://github.com/acornjs/acorn/pull/844

mikesherov:
FYI, I've also pinged @sokra about the change so hopefully it can land in Acorn before Webpack 5.

mikesherov:
More FYI, @sokra has said this can land in Webpack 4, and them doesn't represent a breaking change they need to communicate to their users.

mikesherov:
Just for record keeping:
This has landed in Acorn 7.0.0 and will be in Webpack 5 alpha 21.

PR #205, opened 2019-11-26
Add exported property to ExportAllDeclaration

### LINKED ISSUE #202: Export * as ns from "source" representation
The [proposal](https://spectranaut.github.io/proposal-export-ns-from) to enable the new syntax `export * as ns from "source"` has been merged into EcmaScript standard on master for the 2020 edition.
I don't see any change about it in es202.md, how can i represent it?

mck89:
Closes #202

### DIFF
--- a/es2020.md
+++ b/es2020.md
@@ -47,9 +47,7 @@ interface ImportExpression <: Expression {
 ##  ExportAllDeclaration
 
 ```js
-interface ExportAllDeclaration <: ModuleDeclaration {
-  type: "ExportAllDeclaration";
-  source: Literal;
+extend interface ExportAllDeclaration {
   exported: Identifier | null;
 }
 ```
--- a/es2020.md
+++ b/es2020.md
@@ -41,3 +41,16 @@ interface ImportExpression <: Expression {
   node, but it can be an arbitrary expression node.
 
 [ImportExpression]: es2015.md#importdeclaration
+
+# Modules
+
+##  ExportAllDeclaration
+
+```js
+interface ExportAllDeclaration <: ModuleDeclaration {
+  type: "ExportAllDeclaration";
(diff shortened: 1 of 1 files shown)


es2020.md
```diff
@@ -41,3 +41,16 @@ interface ImportExpression <: Expression {
   node, but it can be an arbitrary expression node.
 
 [ImportExpression]: es2015.md#importdeclaration
+
+# Modules
+
+##  ExportAllDeclaration
+
+```js
+interface ExportAllDeclaration <: ModuleDeclaration {
```
RReverser: If it's extending an existing interface, this should say `extend interface`.

mck89: I've corrected the code as you said

RReverser:
@loganfsmyth What do you think about this vs `ExportNamespaceDeclaration`? (adding to reviewers as you had some concerns)

mysticatea:
For what it's worth, I think that `ExportAllDeclaration` is proper because it takes all members from `"source"` then exports them. The word "all" affects to source rather than destination, so we can write other export declarations even if `ExportAllDeclaration` existed.

mysticatea:
If `ExportAllDeclaration` means "it exports an ambiguous number of elements (in other words, it needs to load more files to decide the number of exported elements)," then using `ExportNamedDeclaration` is reasonable. But `ModuleSpecifier` has `local` property always, we should fix it.

mck89:
Any update about this?

jdalton:
Just fyi this PR is holding up support in projects like `acorn`.

kaicataldo:
One more friendly ping here.

bradzacher:
Friendly ping here, as this is holding up support for TS3.8 features in typescript-eslint (and thus in prettier as well).

PR #209, opened 2020-01-29
Add latest.md, fix typo in es2020.md

### LINKED ISSUE #208: A single specification?
Why is there not a single specification that contains all of the latest AST definitions? Unless I am missing something, feels like a no-brainer to have one up to date self sufficient specification so we don't have to juggle through 7 different documents.

j-f1:
Fixes #208

**[View Rendered File](https://github.com/j-f1/forked-estree/blob/add-latest/latest.md)**


RReverser:
Hmm, we probably need a way to run this automatically on commit, and/or at least verified on CI side.

j-f1:
Agreed, but it would be difficult as we have to merge the declarations

armano2:
maybe `declarations` should be written in structured files like yaml or json and documentation than can be generated

j-f1:
@armano2 https://github.com/estree/formal

RReverser:
Either estree/formal or simply one of the Markdown TOC generators should help.

j-f1:
> Agreed that this should be auto-generated. Can someone explain how `latest.md` was generated for this PR?

Manually! I worked on a different approach which I’m going to PR shortly

j-f1:
See #232 which makes bigger changes to allow automatic generation.

PR #210, opened 2020-04-02
Add `import.meta` to ES2020

mysticatea:
> https://github.com/tc39/proposals/commit/0c27379ebb525ffebed75584827ce51687af99f8

`import.meta` meta property syntax has arrived at Stage 4 in the 75th TC39 meeting.

This PR brings the syntax into ESTree.
The existing [MetaProperty] node represents `import.meta` meta property as well.

[MetaProperty]: https://github.com/estree/estree/blob/master/es2015.md#metaproperty

### DIFF
--- a/es2015.md
+++ b/es2015.md
@@ -303,6 +303,8 @@ interface MetaProperty <: Expression {
 }
 ```
 
+- `MetaProperty` node represents `new.target` meta property in ES2015. In the future, it will represent other meta properties as well.
+
 # Modules
 
 ## ModuleDeclaration
--- a/es2020.md
+++ b/es2020.md
@@ -37,10 +37,12 @@ interface ImportExpression <: Expression {
 ```
 
 - `ImportExpression` node represents Dynamic Imports such as `import(source)`.
-  The `source` property is the importing source as similar to [ImportExpression]
+  The `source` property is the importing source as similar to [ImportDeclaration]
   node, but it can be an arbitrary expression node.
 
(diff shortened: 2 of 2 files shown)


RReverser:
Looks reasonable, although I'm not sure if we need to explicitly list all meta properties (at least we didn't do that in the original PR #32 either).

mysticatea:
@RReverser I think worthful if we mention it explicitly because ESTree was representing dynamic imports and `import.meta` with `Import` node in the past (https://github.com/estree/estree/pull/137).

RReverser:
@mysticatea Huh. Dynamic imports, sure, but `import.meta`? If it did, it probably was a bug in a parser (or an accident), not something intentional. The PR you referred only mentions dynamic import.

Anyway, adding explicit mention of meta property sounds good, but can you please add the existing `new.target` mention to es5.md too then for consistency?

mysticatea:
Sure, I added the mention of `new.target`.

<details><summary>Off-topic:</summary>

(in my understanding, the `Import` node was introduced to represent `import()` and `import.foo` in the same manner of `Super`. So, there was a question about `import.meta` when I brought `ImportExpression` node (https://github.com/estree/estree/issues/197).)

</details>

RReverser:
> in my understanding, the Import node was introduced

Ah, I see what you're saying.

We did temporarily have `Import` in an experimental folder, but we ended up with `ImportExpression` instead, which doesn't (can't) cover `import.meta`, hence the confusion.

PR #204, opened 2019-11-20
Add optional chaining

mysticatea:
This is a draft to update `es2020.md` for the TC39 meeting in Dec 2019. The meeting contains the following items on its agenda (https://github.com/tc39/agendas/blob/master/2019/12.md).

- Optional Chaining for Stage 4
- Nullish Coalescing for Stage 4

Therefore, this PR contains the AST proposal of those three syntaxes. If any of them didn't arrive at Stage 4, I will remove those from this PR.

(Because we needed over 2 months to merge a PR in the previous time, I wanted to start discussion earlier.)

----

**Optional Chaining:**

This proposal is inspired by https://github.com/estree/estree/issues/146#issuecomment-365781075. It adds two properties `optional` and `shortCircuited` to two existing nodes `CallExpression` and `MemberExpression`.

The following examples describe how the two properties represent the syntax.

<details>

In the following code blocks, the first comment is the source code and the rest is the AST of the code.

```jsonc
// obj?.aaa?.bbb
{
  "type": "MemberExpression",
  "optional": true,     // If `obj.aaa` was nullish, this node is evaluated to undefined.
  "shortCircuited": true, // If `obj` was nullish, this node is short-circuited.
  "object": {
    "type": "MemberExpression",
    "optional": true,      // If `obj` was nullish, this node is evaluated to undefined.
    "shortCircuited": false, // This node isn't short-circuited.
    "object": { "type": "Identifier", "name": "obj" },
    "property": { "type": "Identifier", "name": "aaa" },
  },
  "property": { "type": "Identifier", "name": "bbb" },
}
```

```jsonc
// obj?.aaa.bbb
{
  "type": "MemberExpression",
  "optional": false,    // If `obj.aaa` was nullish, this node is evaluated to throwing TypeError.
  "shortCircuited": true, // If `obj` was nullish, this node is short-circuited.
  "object": {
    "type": "MemberExpression",
    "optional": true,      // If `obj` was nullish, this node is evaluated to undefined.
    "shortCircuited": false, // This node isn't short-circuited.
    "object": { "type": "Identifier", "name": "obj" },
    "property": { "type": "Identifier", "name": "aaa" },
  },
  "property": { "type": "Identifier", "name": "bbb" },
}
```

```jsonc
// (obj?.aaa)?.bbb
{
  "type": "MemberExpression",
  "optional": true,      // If `obj.aaa` was nullish, this node is evaluated to undefined.
  "shortCircuited": false, // This node isn't short-circuited.
  "object": {
[...]

PR #212, opened 2020-04-17
add Logical Assignment Operators

mysticatea:
Recently, [Logical Assignment Operators][proposal-logical-assignment] arrived at Stage 3.

Considering AST before Stage 4 is nice.

[proposal-logical-assignment]: https://github.com/tc39/proposal-logical-assignment

### DIFF
--- a/experimental/logical-assignment-operators.md
+++ b/experimental/logical-assignment-operators.md
@@ -2,29 +2,16 @@
 
 ## Expressions
 
-### LogicalAssignmentExpression
+### AssignmentOperator
 
 ```js
-interface LogicalAssignmentExpression <: Expression {
-    type: "LogicalAssignmentExpression";
-    operator: LogicalAssignmentOperator;
-    left: Identifier | MemberExpression;
-    right: Expression;
-}
-```
-
-- The relationship of `AssignmentExpression` and `LogicalAssignmentExpression`
-  is similar to the relationship of `BinaryExpression` and `LogicalExpression`.
-  The `LogicalAssignmentExpression` node has short-circuit behavior -- it skips
-  the evaluation of the `right` property by the evaluated value of the `left`
-  property.
-
-### LogicalAssignmentOperator
-
-```js
-enum LogicalAssignmentOperator {
+extend enum AssignmentOperator {
     "||=" | "&&=" | "??="
 }
 ```
 
+- [AssignmentExpression] node has short-circuiting behavior if the `operator`
+  property is any of `"||="`,`"&&="`, and `"??="`.
+
 [proposal-logical-assignment]: https://github.com/tc39/proposal-logical-assignment
+[AssignmentExpression]: ../es5.md#AssignmentExpression
--- a/experimental/logical-assignment-operators.md
+++ b/experimental/logical-assignment-operators.md
@@ -8,7 +8,7 @@
 interface LogicalAssignmentExpression <: Expression {
     type: "LogicalAssignmentExpression";
     operator: LogicalAssignmentOperator;
-    left: Pattern | Expression;
+    left: Identifier | MemberExpression;
     right: Expression;
 }
 ```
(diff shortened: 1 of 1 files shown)


experimental/logical-assignment-operators.md
```diff
@@ -0,0 +1,30 @@
+# [Logical Assignment Operators][proposal-logical-assignment]
+
+## Expressions
+
+### LogicalAssignmentExpression
+
+```js
+interface LogicalAssignmentExpression <: Expression {
```
RReverser, line 8: Why do we need a new node type for this rather than extending `AssignmentOperator` in [`AssignmentExpression`](https://github.com/estree/estree/blob/master/es5.md#assignmentexpression)?

RReverser: Ah, sorry, I see it's explained below.

nzakas: Just to make sure I understand this: a new node type is necessary because logical assignment expressions are short-circuited?

RReverser: As far as I understood, it's for symmetry with `BinaryExpression` vs `LogicalExpression` (which were initially separate due to short-circuiting, I suppose).

Personally, I'd rather have former to be a single node type as well, but that ship has sailed.

As for new node types, I don't have a strong opinion and fine with either a new node type or reusing `AssignmentExpression`. Would like to hear opinions of others.

mysticatea: Yes, it's for symmetry with `BinaryExpression` vs `LogicalExpression`, and I thought that's because of short-circuit.

bradzacher: a question here from my relative lack of knowledge of the history of ESTree, and of how ASTs should be designed.

Why is it that ESTree wants to encode this runtime behaviour in the AST?
Isn't the short-circuiting behaviour semantic meaning derived from the operator?

Is there tangible value for consumers of the spec with this distinction?

I ask because, for example, TypeScript's AST representation just uses a single `BinaryExpression` node for both cases.

RReverser: @bradzacher Yeah, as I said above

> Personally, I'd rather have former to be a single node type as well, but that ship has sailed.

This has been there from the beginning of SpiderMonkey AST and I can only assume that it was for encoding of short-circuiting behaviour, but personally I'd also just use a single node type and encourage consumers to check the operator.

Not sure whether to follow the precedent or do something simpler for the new node types here. *Side-note: this is why I said I'm very cautious of accidentally setting new precedents in #204 - they can affect design of nodes far away from the original addition.*

experimental/logical-assignment-operators.md
```diff
@@ -0,0 +1,30 @@
...
+    type: "LogicalAssignmentExpression";
+    operator: LogicalAssignmentOperator;
+    left: Pattern | Expression;
```
JLHwung, line 11: I don't think `Pattern` is allowed here.

jridgewell: Correct.

mysticatea: Nice catch.

experimental/logical-assignment-operators.md
mysticatea: I don't have any strong opinion about the choice of following the precedent or using `AssignmentExpression` simply.

nzakas: This is why I asked, as I’m not sure if we want to encode short-circuiting operations separately. My preference would be to reuse AssignmentExpression. If the intent is to execute based on the AST, you’ll always need to look at the operator anyway, and that seems like the logical time to apply any short circuiting behavior.

mysticatea: I'm OK with either, but I hope consistent with #204. If we decide to not make new nodes for short-circuiting, I think it's better if #204 is consistent with that.

mysticatea: I have updated this PR to extending `AssignmentOperator` because the separation of nodes by short-circuiting looks not important to ESTree.

nzakas:
We have approvals from all ESC members on this pull request, so it is approved and so I’m merging.

PR #215, opened 2020-04-30
add module-attributes

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/add-module-attributes/experimental/module-attributes.md)

Add stage-1 [Module Attributes proposal](https://github.com/tc39/proposal-module-attributes). The `ImportDeclaration` and `ImportAttributes` part also shows how Babel _plans_ to support this feature.

Since many proposals are experimented in Babel in an earlier stage, I think generally we should more proactively get involved in the experimental proposals so we can have more AST shape feedback before it is implemented in Babel. I will add more experimental proposals later.

### DIFF
--- a/experimental/module-attributes.md
+++ b/experimental/module-attributes.md
@@ -1,0 +1,41 @@
+# [Module Attributes][proposal-module-attributes]
+
+## Imports
+
+### ImportDeclaration
+
+```js
+extend interface ImportDeclaration {
+    attributes: [ ImportAttribute ];
+}
+```
+
+The `attributes` is non-empty when import attributes present, e.g., `import foo from "./foo.json" with type: "json"`
+
+### ImportAttribute
+
+```js
+interface ImportAttribute <: Node {
+    type: "ImportAttribute";
+    key: Identifier;
+    value: Literal;
+}
+```
+
+An import attribute is an object-like key value pair, e.g., `type: "json"` in `import foo from "./foo.json" with type: "json"`. The `value` must be a string literal, that said, `value.value` is always `string`-type.
+
+
+## Expressions
+
+### ImportExpression
+
+```js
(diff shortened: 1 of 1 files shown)


nzakas:
We should probably discuss how we want to handle proposals that are at such an early stage. Given that the syntax could change before finalized, it seems a bit difficult to nail anything down.

How do you envision we should evaluate a stage 1 proposal?

RReverser:
> We should probably discuss how we want to handle proposals that are at such an early stage. Given that the syntax could change before finalized, it seems a bit difficult to nail anything down.

We traditionally have put them into `experimental` folder, just like this PR does. It basically doesn't have any stability guarantees at this stage, just like the actual ECMAScript proposal, and only serves as a PoC for 3rd-party plugins to experiment with.

hzoo:
Yeah that's what current readme has at least https://github.com/estree/estree/blob/master/experimental/README.md - sorta like a first pass to get things going.

nzakas:
Thanks for explaining. The part I’m still unclear on is the expectations of review on a PR for experimental syntax. Is it just a cursory review? Is there any review, or do we just merge and worry about it later? Is it just a repository for Babel to document the first implementation?

RReverser:
> Thanks for explaining. The part I’m still unclear on is the expectations of review on a PR for experimental syntax. Is it just a cursory review?

We still look for obvious things that would never get merged (e.g. break backwards compatibility) or invalid representations, but just avoid having controversial discussions if any arise.

nzakas:
Sounds good.

JLHwung:
It seems we reach consensus on this PR: I can't approve my own PR.

Does anyone object?

RReverser:
Just merged, thanks!

PR #216, opened 2020-05-05
add wavy dot

JLHwung:
This PR depends on #204.

The draft extends from the `ChainingExpression` approach proposed by the ESLint team. I choose this approach as it is more accessible the the other. I will work the other if we decide to use the `OptionalCallExpression` approach.

Since `a~.b()` and `(a~.b)()` have different semantics, we add `eventual: boolean` to `Chain` instead of `(Member|Call)Expression`. I think this PR can offer perspective on how our decision on #204 will affect other proposal designs.

Note that although the examples contain `?~.`, it should be considered as an experiment of an indeed experimental proposal which has not supported optional chaining yet. Although the optional chaining support can be expected in the near future, I have added a clause that `eventual` and `optional` cannot be both true to align to current proposal.

### DIFF
--- a/experimental/wavy-dot.md
+++ b/experimental/wavy-dot.md
@@ -1,0 +1,17 @@
+# [Wavy Dot](https://github.com/tc39/proposal-wavy-dot)
+
+Extend from the `ChainingExpression` [approach](https://gist.github.com/mysticatea/f3a87f3e02632797ec59d9b447fdf05e).
+
+# Expressions
+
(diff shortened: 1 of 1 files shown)


RReverser:
I think this needs to be simply updated to use `ChainElement`.

michaelfig:
> I think this needs to be simply updated to use `ChainElement`.

I'd like to pick this up again.  Can you please clarify?

PR #217, opened 2020-05-09
add private-fields-in-in

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/add-private-fields-in-in/experimental/private-fields-in-in.md)

Added stage-1 [Ergonomic brand checks for Private Fields proposal](https://github.com/tc39/proposal-private-fields-in-in).

This PR depends on #180 where `PrivateName` is defined. Babel also adopts this approach.

### DIFF
--- a/experimental/private-fields-in-in.md
+++ b/experimental/private-fields-in-in.md
@@ -5,7 +5,7 @@
 ### BinaryExpression
 
 ```js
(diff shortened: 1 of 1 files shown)


nzakas:
@rreverser just waiting for your :+1: here

PR #218, opened 2020-05-09
add record-tuple

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/add-record-tuple/experimental/record-tuple.md)

This PR adds stage-1 [Record & Tuple](https://github.com/tc39/proposal-record-tuple) support. It is a close variant of Babel's current implementation. (Babel uses `ObjectProperty` instead)

### Open question
- Should we merge `RecordExpression` and `ObjectExpression` by adding an extra `immutable: boolean` property to `ObjectExpression`? Likewise for `TupleExpression`.

### DIFF
--- a/experimental/record-tuple.md
+++ b/experimental/record-tuple.md
@@ -1,0 +1,28 @@
+# [Record & Tuple][proposal-record-tuple]
+
+## Expressions
+
+### RecordExpression
+
+```js
+interface RecordExpression <: Expression {
+    type: "RecordExpression";
+    properties: [ Property | SpreadElement ];
+}
+```
+
+A record expression defines an immutable object, e.g. `#{a: 1}`.
+
+For each element _p_ of `properties`, if _p_ is a `Property`, `p.method` must be `false` and  `p.kind` must be `"init"`.
+
+### TupleExpression
+```js
+interface TupleExpression <: Expression {
+    type: "TupleExpression";
+    elements: [ Expression | SpreadElement ];
+}
+```
+
+A tuple expression defines an immutable array, e.g. `#[1, 2]`. Holes are disallowed in tuple expressions.
+
+[proposal-record-tuple]: https://github.com/tc39/proposal-record-tuple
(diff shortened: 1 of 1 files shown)


nzakas:
@rreverser thoughts?

experimental/record-tuple.md
```diff
@@ -0,0 +1,28 @@
+# [Record & Tuple][proposal-record-tuple]
+
+## Expressions
+
+### RecordExpression
+
+```js
+interface RecordExpression <: Expression {
+    type: "RecordExpression";
+    properties: [ Property | SpreadElement ];
```
RReverser, line 10: Hmm it might be a bit early to go into details, so I suppose we could go as-is for now, but my understanding is that records will be more limited (e.g. no getters/setters/methods, right?). If so, we might want to have a separate RecordProperty node down the road.

JLHwung: Good catch. In Babel we have `ObjectMethod` and `ObjectProperty` node, I just realize that they are both `Property` in estree. I am good to introduce a `RecordProperty`, will update this PR later.

experimental/record-tuple.md
```diff
@@ -0,0 +1,37 @@
...
+    type: "RecordExpression";
+    properties: [ RecordProperty | SpreadElement ];
+}
+```
+
+A record expression defines an immutable object, e.g. `#{a: 1}`.
+
+#### RecordProperty
+```js
+interface RecordProperty <: Node {
```
RReverser, line 18: We'll probably want to extract a `BaseProperty` mixin shared by both kinds of properties (object and record), but we can resolve this later if/when proposal is accepted, since it's a readability nit that doesn't change anything semantically.

RReverser: (Of course, another alternative here could be to keep the `"Property"` node type, but override `kind` to strictly `"init"` like we do on some other subtypes.)

JLHwung: I like the idea of `BaseProperty` mixin! Note that we can also share it with `ClassProperty` or `ClassElement` defined in #180. Since it does not change semantic we can refactor in a separate PR when #180 is resolved.

PR #220, opened 2020-05-12
add decimal

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/add-decimal/experimental/decimal.md)

This PR adds support on stage-1 [Decimal Proposal](https://github.com/tc39/proposal-decimal).

Note that as of May 2020, TC39 hasn't decided between two competing decimal implementations: `bigdecimal` and `decimal128`. This PR adopts the `bigdecimal` approach. Note that if in the future `decimal128` is chosen instead, we can replace the type annotation of `value` by `decimal128`.

Other than that, whether `bigdecimal` or `decimal128` is chosen should not affect the AST shapes and the new interface name.

### DIFF
--- a/experimental/decimal.md
+++ b/experimental/decimal.md
@@ -1,0 +1,25 @@
+# [Decimal][proposal-decimal]
+
+# Literal
+
+```js
+extend interface Literal <: Expression {
+  type: "Literal";
+  value: string | boolean | null | number | RegExp | bigint | bigdecimal
+}
+```
+
+- `value` property can be a `bigdecimal` value to represent decimal literals, e.g. `2.718m`.
+
+## DecimalLiteral
+
+```js
+interface DecimalLiteral <: Literal {
+  decimal: string;
+}
+```
+
+- `decimal` property is the string representation of the `bigdecimal` value. It doesn't include the suffix `m`.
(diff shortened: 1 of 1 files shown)


experimental/decimal.md
```diff
@@ -0,0 +1,25 @@
...
+```js
+extend interface Literal <: Expression {
+  type: "Literal";
+  value: string | boolean | null | number | RegExp | bigint | bigdecimal
+}
+```
+
+- `value` property can be a `bigdecimal` value to represent decimal literals, e.g. `2.718m`.
+
+## DecimalLiteral
```
nzakas, line 14: In ESTree we haven't had separate nodes for different types of literals. I know Babel does, but maybe it should be left out here?

JLHwung: Note that the `DecimalLiteral` does not have a "DecimalLiteral" type property. For `0.3m`, it has an extra `decimal: string` property that stores the string representation `0.3`.

We have similar settings for `BigIntLiteral`: https://github.com/estree/estree/blob/master/es2020.md#bigintliteral

nzakas: Oh I see. Thanks for explaining.

PR #222, opened 2020-05-28
add class features

### LINKED ISSUE #154: Private Class Field representation
Some discussion in https://github.com/babel/babylon/pull/260 currently, figured I'd raise it over here before we accept anything.

Current proposal there being
```
interface PrivateName <: Expression, Pattern {
  type: "PrivateName";
  name: string;
}

interface ClassPrivateProperty <: Node {
  type: "ClassPrivateProperty";
  key: Identifier;
  value: Expression;
}

extend interface ClassBody {
  body: [ MethodDefinition | ClassPrivateProperty ]
}
```

I'm not sure I love the duplication between `PrivateName` and `Identifier` so perhaps `name` could be an `Identifier` instead of a string.

Keep in mind that both `this.#priv = 1234` and `#priv = 1234` are allowed, otherwise there'd be the potential for `MemberExpression` to just have a `private: boolean` property.

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/add-class-features/experimental/class-features.md)

This PR adds support to three stage-3 class features proposals. It shows Babel's current approach except for the following node type naming differences:

| Babel | This PR |
| --- | --- |
| ClassProperty | PropertyDefinition |
| ClassPrivateProperty | PrivatePropertyDefinition |
| ClassPrivateMethod | PrivateMethodDefinition |

It is named in the `*Definition` pattern, aligning to the `MethodDefinition` defined in `es2015`.

This PR also offers an alternative design of #180. In #180 `PrivateMemberExpression` has been introduced for `MemberExpression[property=PrivateName]` in this PR, which will have impact on #204 given that `this?.#priv` is now supported.

As the class proposals have been out there for quite some time and they are widely adopted in React/TypeScript communities, especially the class fields proposal, I will expect some tools are depending on Babel class features AST. Building consensus based on this approach can minimize the friction supporting new ESTree releases.

Fixes #154
Closes #180  

--- Edits ---
Updated according to review comments. PTAL.

### DIFF
--- a/experimental/class-features.md
+++ b/experimental/class-features.md
@@ -7,7 +7,7 @@ This experimental extension covers three class features proposals:
 ## ClassBody
 
 ```js
-extend interface ClassBody <: Node {
+extend interface ClassBody {
     body: [ MethodDefinition | PropertyDefinition ];
 }
 ```
@@ -33,14 +33,12 @@ interface PropertyDefinition <: Node {
 ```js
 extend interface MethodDefinition {
     key: Expression | PrivateName;
-    static: boolean;
     private: boolean;
 }
 ```
 
 - When `private` is `true`, `computed` must be `false`, `key` must be a `PrivateName`, `kind` can not be `"constructor"`.
 - When `private` is `false`, `key` must be an `Expression` and can not be a `PrivateName`.
-- When `static` is `true`, `kind` can not be `"constructor"`.
 
 ### PrivateName
 
@@ -54,7 +52,7 @@ interface PrivateName <: Node {
 A private name refers to private class elements. For a private name `#a`, its `name` is `a`.
 
 ```js
-extend interface MemberExpression <: ChainElement {
+extend interface MemberExpression {
     private: boolean;
     property: Expression | PrivateName;
 }
--- a/experimental/class-features.md
+++ b/experimental/class-features.md
@@ -4,40 +4,42 @@
 This experimental extension covers three class features proposals: 
 [Class Fields], [Static Class Features] and [Private Methods].
 
+## ClassBody
+
+```js
+extend interface ClassBody <: Node {
+    body: [ MethodDefinition | PropertyDefinition ];
+}
+```
+
 ## PropertyDefinition
 
 ```js
 interface PropertyDefinition <: Node {
     type: "PropertyDefinition";
-    key: Expression;
+    key: Expression | PrivateName;
     value: Expression;
     computed: boolean;
     static: boolean;
+    private: boolean;
 }
 ```
 
-## PrivatePropertyDefinition
+- When `private` is `true`, `computed` must be `false` and `key` must be a `PrivateName`.
+- When `private` is `false`, `key` must be an `Expression` and can not be a `PrivateName`.
+
+## MethodDefinition
 
 ```js
-interface PrivatePropertyDefinition <: Node {
-    type: "PrivatePropertyDefinition";
-    key: PrivateName;
-    value: Expression;
+extend interface MethodDefinition <: Node {
+    key: Expression | PrivateName;
     static: boolean;
+    private: boolean;
 }
 ```
 
-## PrivateMethodDefinition
-
-```js
-interface PrivateMethodDefinition <: Node {
-    type: "PrivateMethodDefinition";
-    key: PrivateName;
-    value: FunctionExpression;
-    kind: "method" | "get" | "set";
-    static: boolean;
-}
-```
+- When `private` is `true`, `computed` must be `false`, `key` must be a `PrivateName`, `kind` can not be `constructor`.
+- When `private` is `false`, `key` must be an `Expression` and can not be a `PrivateName`.
 
 ### PrivateName
 
@@ -54,7 +56,7 @@ extend interface MemberExpression {
 
 A private name refers to private class elements. For a private name `#a`, its `id.name` is `a`.
 
-When the property of a member expression is a private name, its `computed` is always `false`.
+When the `property` of a `MemberExpression` is a `PrivateName`, `computed` is always `false`.
 
 [Class Fields]: https://github.com/tc39/proposal-class-fields
 [Static Class Features]: https://github.com/tc39/proposal-static-class-features/
--- a/experimental/class-features.md
+++ b/experimental/class-features.md
@@ -48,15 +48,19 @@ interface PrivateName <: Node {
     type: "PrivateName";
     id: Identifier;
 }
+```
 
-extend interface MemberExpression {
-    property: Expression | PrivateName
+```js
+extend interface MemberExpression <: ChainElement {
+    private: boolean;
+    property: Expression | PrivateName;
 }
 ```
 
 A private name refers to private class elements. For a private name `#a`, its `id.name` is `a`.
 
-When the `property` of a `MemberExpression` is a `PrivateName`, `computed` is always `false`.
+- When `private` is `true`, `property` must be a `PrivateName`, `computed` must be `false`.
+- When `object` is a `Super`, `private` must be `false`.
 
 [Class Fields]: https://github.com/tc39/proposal-class-fields
 [Static Class Features]: https://github.com/tc39/proposal-static-class-features/
--- a/experimental/class-features.md
+++ b/experimental/class-features.md
@@ -17,7 +17,7 @@ extend interface ClassBody {
 ```js
 interface PropertyDefinition <: Node {
     type: "PropertyDefinition";
-    key: Expression | PrivateName;
+    key: Expression | PrivateIdentifier;
     value: Expression | null;
     computed: boolean;
     static: boolean;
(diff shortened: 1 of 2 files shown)


mysticatea:
I wonder if the `Private*` prefix is needed or not. I love consistency.

- `MemberExpression` handles both public and private.
- There are four kinds of namespaces: public keys, static public keys, private keys, and static private keys. `MethodDefinition` doesn't have `Static*` prefix for static keys. (also I'm afraid of the combination explosion of such prefixes.)

experimental/class-features.md
```diff
@@ -0,0 +1,61 @@
...
+}
+
+extend interface MemberExpression {
+    property: Expression | PrivateName
+}
+```
+
+A private name refers to private class elements. For a private name `#a`, its `id.name` is `a`.
+
+When the property of a member expression is a private name, its `computed` is always `false`.
```
michaelficarra, line 57: This seems pretty bad. Why not just have a `PrivateMemberExpression`? The reason to unify them would be that we expect people to consider them similar for some purpose. I don't expect too much of that.

JLHwung: Some reasons that I can tell

- Babel has `OptionalMemberExpression` which corresponds to [OptionalChain](https://tc39.es/ecma262/#prod-OptionalChain)
- `PrivateOptionalMemberExpression` for `obj?.#x` can be confusing
- There is no `PrivateMemberExpression` like production in the spec

experimental/class-features.md
```diff
@@ -0,0 +1,63 @@
...
+
+- When `private` is `true`, `computed` must be `false` and `key` must be a `PrivateName`.
+- When `private` is `false`, `key` must be an `Expression` and can not be a `PrivateName`.
+
+## MethodDefinition
+
+```js
+extend interface MethodDefinition <: Node {
+    key: Expression | PrivateName;
+    static: boolean;
```
JLHwung, line 36: The `static: boolean` property is also available in https://github.com/estree/estree/blob/master/es2015.md#methoddefinition

I think this is a mistake and `MethodDefinition ` in `es2015.md` should not have such a property, although acorn has supported it.

experimental/class-features.md
```diff
@@ -0,0 +1,63 @@
...
+}
+```
+
+## PropertyDefinition
+
+```js
+interface PropertyDefinition <: Node {
+    type: "PropertyDefinition";
+    key: Expression | PrivateName;
+    value: Expression;
```
mysticatea, line 21: The initializer of [FieldDefinition](https://tc39.es/proposal-class-fields/#sec-new-syntax) production is optional. So should the `value` nullable?

experimental/class-features.md
```diff
@@ -0,0 +1,63 @@
...
+
+- When `private` is `true`, `computed` must be `false`, `key` must be a `PrivateName`, `kind` can not be `constructor`.
+- When `private` is `false`, `key` must be an `Expression` and can not be a `PrivateName`.
+
+### PrivateName
+
+```js
+interface PrivateName <: Node {
+    type: "PrivateName";
+    id: Identifier;
```
mysticatea, line 49: How about just `name: string`? Because the [PrivateIdentifier](https://tc39.es/proposal-class-fields/#prod-PrivateIdentifier) production is a token that we cannot split it anymore, I feel odd using the `Identifier` node here.

bradzacher:
I understand the want to align with `MethodDefinition`, but I question if it's a good idea to rename `ClassProperty` to `PropertyDefinition`. The `ClassProperty` name has existed for quite a long time, and as such it has been used for a long time. It is very, very prolific throughout the ecosystem.

Unlike the optional chaining changes, which apply to a feature that has only recently been implemented in most of the ecosystem, renaming class properties will be a pretty significant breaking change and will likely be a pretty painful change for users to navigate.

experimental/class-features.md
```diff
@@ -0,0 +1,63 @@
...
+## PropertyDefinition
+
+```js
+interface PropertyDefinition <: Node {
+    type: "PropertyDefinition";
+    key: Expression | PrivateName;
+    value: Expression;
+    computed: boolean;
+    static: boolean;
+    private: boolean;
```
mysticatea, line 24: I love consistency. If we add `private:boolean` to `(Method|Property)Definition`, I'd like to add it to `MemberExpression` as well. Or remove it from both.

experimental/class-features.md
JLHwung: Related: https://github.com/babel/babylon/pull/260#pullrequestreview-38805907

JLHwung:
@bradzacher The `ClassProperty` is used in Babel AST. Given that `PropertyDefinition` now deviates a lot from Babel AST design. I would rather stay on the current names instead of silently breaking tools. (i.e. `private: boolean` is added)

experimental/class-features.md
RReverser: Why do you think it's a mistake? Static methods were available in ES2015.

experimental/class-features.md
```diff
@@ -0,0 +1,69 @@
...
+}
+```
+
+- When `private` is `true`, `computed` must be `false` and `key` must be a `PrivateName`.
+- When `private` is `false`, `key` must be an `Expression` and can not be a `PrivateName`.
+
+## MethodDefinition
+
+```js
+extend interface MethodDefinition <: Node {
```
RReverser, line 34:
```suggestion
extend interface MethodDefinition {
```

experimental/class-features.md
RReverser: I think it's best to remove from both; otherwise we're duplicating same information which explicitly goes against one of the design goeals ("Unique") in the Philosophy.

experimental/class-features.md
JLHwung: Oh! My bad! I thought it is introduced in the static proposal, will update it.

bradzacher:
> I would rather stay on the current names instead of silently breaking tools.

Would it be a good idea to update this PR to match that, so that it doesn't get committed with the new names?
I wouldn't want to have it committed and then have some new use cases built on top of the new experimental names and then have the ecosystem in two states.

experimental/class-features.md
```diff
@@ -0,0 +1,67 @@
...
+}
+```
+
+- When `private` is `true`, `computed` must be `false`, `key` must be a `PrivateName`, `kind` can not be `"constructor"`.
+- When `private` is `false`, `key` must be an `Expression` and can not be a `PrivateName`.
+
+### PrivateName
+
+```js
+interface PrivateName <: Node {
```
mysticatea, line 46: Maybe we should use a different name because Babel's AST has a node which is the same name but a different shape. 
 How about [PrivateIdentifier](https://tc39.es/proposal-class-fields/#prod-PrivateIdentifier) as same as the language spec's token name?

RReverser: `PrivateIdentifier` sounds good to me.

experimental/class-features.md
JLHwung: I agree on @RReverser that we don't need `.private` if we can check `.key.type === "PrivateIdentifier"`. While I admit that `.private` is handier, the ergonomic trade-off here is still acceptable.

Should another private-like key (`PrivateComputedExpression`?) be supported in the future, we can introduce `.private` at that time.

JLHwung:
@RReverser friendly ping, does the current AST design look good to you?

RReverser:
Oh yeah, sorry, I haven't noticed that `private: boolean` has been removed since my last review; LGTM, especially for experimental :)

bradzacher:
> Given that PropertyDefinition now deviates a lot from Babel AST design

Sorry, could you clarify how? I don't see how this deviates from the babel AST (or the AST emitted from `babel-eslint`).

It looks like the current `ClassProperty` spec introduced by babel and used in the ESLint ecosystem for many years matches your proposed spec exactly (with the only variance being the private property piece)

AST explorer example:
https://astexplorer.net/#/gist/592095bc625a5243ab44009fd1d62863/15d212b3d4439339de5fd36cb0b057d293d0d003

TS-ESTree spec (which was built to match the babel implementation):
https://github.com/typescript-eslint/typescript-eslint/blob/4c9293bd5ecfb475bfd4f2240fb6e55186ff3e55/packages/typescript-estree/src/ts-estree/ts-estree.ts#L583-L606

The `ClassProperty` name has been around and in use for many, many years, so this will cause a lot of pain for tool owners and users to migrate to.

If that's what the direction going forward is - that's fine. I just want to ensure it's a conscious decision to introduce the breakage.

JLHwung:
The differences between current approach and Babel's approach can be viewed [here](https://github.com/JLHwung/estree/compare/5c442f18921cb04a76e4e5b196b15ba1cda38bd0...JLHwung:add-class-features)

Notable changes are:

- `ClassPrivateProperty` and `ClassProperty` is merged, likewise `ClassPrivateMethod`
- Naming differences
- `PrivateName` has been replaced by `PrivateIdentifier`

On the AST policy of `babel` and `estree`, we haven't stated formally on our website but I believe what @nicolo-ribaudo proposed [here](https://github.com/babel/babel/issues/9506#issuecomment-564926566) is close to the team consensus, that `estree` does not include experimental features and Babel's proposal AST design should be considered experimental, too. That said, we as Babel team don't intend to bring more fragments to the ECMAScript AST, that's why we have proposed an [update](https://github.com/babel/babel/pull/11706) of our process to work with estree since stage-1.

Babel introduced `ClassPrivateProperty` in https://github.com/babel/babylon/pull/260 and back that time there have been discussions about naming conventions. The [decision](https://github.com/babel/babylon/pull/260#issuecomment-301673777) at that time was to merge `ClassPrivateProperty` and `ClassProperty` when the class proposals reached stage-3. Now that they are stage-3 and we may consider merging `ClassPrivateProperty` and `ClassProperty`.

> The ClassProperty name has been around and in use for many, many years, so this will cause a lot of pain for tool owners and users to migrate to.

Yeah I understand the frictions here, but keep in mind that class proposals are still stage-3, which means they are still experimental, and by nature, open to breaking changes. It's a short time window that we still have opportunities to value long term AST consistency/teachability over short term community frictions.

JLHwung:
@nzakas I think we have reached consensus and it can merged as-is.

nzakas:
Sounds good!

piranna:
In what Babel version will this be available?

JLHwung:
@piranna Babel has different AST shape with ESTree: https://github.com/babel/babel/blob/main/packages/babel-parser/ast/spec.md#classproperty

All stage-3 class features are supported on Babel 7.10.

piranna:
> @piranna Babel has different AST shape with ESTree: [babel/babel:packages/babel-parser/ast/spec.md@`main`#classproperty](https://github.com/babel/babel/blob/main/packages/babel-parser/ast/spec.md?rgh-link-date=2020-07-07T21%3A00%3A32Z#classproperty)
> 
> All stage-3 class features are supported on Babel 7.10.

Sorry, I wanted to say `eslint`.

PR #223, opened 2020-05-29
add do-expressions

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/do-expressions/experimental/do-expressions.md)

This PR adds support to [Do expressions] proposal. It reflects Babel's current approach.

[Do expressions]: https://github.com/tc39/proposal-do-expressions/

### DIFF
--- a/experimental/do-expressions.md
+++ b/experimental/do-expressions.md
@@ -1,0 +1,14 @@
+# [Do expressions][proposal-do-expressions]
+
+## Expressions
+
+### DoExpression
+
+```js
+interface DoExpression <: Expression {
+    type: "DoExpression";
+    body: BlockStatement;
+}
+```
+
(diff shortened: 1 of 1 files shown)


RReverser:
I thought these were pretty much abandoned - do we want to document them still?

michaelficarra:
@RReverser They're on next week's TC39 agenda, so not entirely abandoned.

RReverser:
Ah great. Just no activity on the repo, which seems surprising if it's still developed.

JLHwung:
@RReverser I think we should wait for approval from @nzakas.

RReverser:
Maybe... We probably should write down somewhere on what we expect on experimental proposals after all. Personally, I don't mind them being merged with one review / without waiting for me if someone already took a look, but encoding that somewhere would be good.

Issue #219, opened 2020-05-12
Define scope of this project


nzakas:
To aid in communicating with participants, it would be helpful to come up with a statement of scope so everyone knows what is and is not a part of ESTree. There has traditionally been some confusion around this:

* [Do we want to define expectations around source locations?](https://github.com/estree/estree/issues/80)
* [Do we want to incorporate a concrete syntax tree?](https://github.com/estree/estree/issues/41)
* [Do we want to include common extensions like JSX?](https://github.com/estree/estree/issues/44)
* [Do we want to include comment types?](https://github.com/estree/estree/issues/201)

From the response (or lack thereof) to these various issues, my guess is the answer to each is "no," but it would be nice to get some consensus and then narrowly define the scope of the project so we can more easily respond to issues and PRs as they come up.

michaelficarra:
> Do we want to define expectations around source locations?

I think this is reasonable. I agree with what @RReverser expressed in the linked thread, that we should have an agreement on what portion of source text is being represented by an AST node, regardless of how concrete syntax information is being conveyed.

> Do we want to incorporate a concrete syntax tree?

There's many ways to convey concrete syntax information, and each has its benefits and drawbacks depending on the use case. I don't think there should be single blessed way to represent and pass around concrete syntax info. For most cases, I prefer a side table style approach, but I can also appreciate the convenience (and difficulties) of inlining concrete syntax info into the AST.

> Do we want to include common extensions like JSX?

Please no. The format is extensible, let them extend it. The most I could see the estree project doing is coordinating reserved prefixes or something for type names to avoid extensions stepping on each others' toes.

> Do we want to include comment types?

No, for the same reason we don't include concrete syntax information like whitespace, position, and unnecessary grouping. Comment representations are also varied and have the same tradeoffs I mentioned with concrete info above. I'd be interested to see either this project or a similar project try to standardise the various comment "attachment" rules, though, defining how comments associate with adjacent nodes. In our [comment-based templating system](https://github.com/shapesecurity/shift-template-js), we ran in to some tricky cases and disambiguated them with optional node types included in the comment.

RReverser:
I agree that answer is "no" in general to most of these questions.

If we ever have enough people and interest to tackle problem of CST again, we could create it as a separate project under the `estree` org and design it in a way compatible with ESTree.

[...]

PR #224, opened 2020-05-30
add hashbang support

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/hashbang/experimental/hashbang.md)

This PR adds stage-3 [Hashbang Grammar] support. It reflects on Babel's current approach.

[Hashbang Grammar]: https://github.com/tc39/proposal-hashbang

### DIFF
--- a/experimental/hashbang.md
+++ b/experimental/hashbang.md
@@ -1,0 +1,22 @@
+# [Hashbang Grammar]
+
+## Programs
+
+```js
+extend interface Program {
+    interpreter: InterpreterDirective | null
+}
+```
+
+### InterpreterDirective
+
+```js
+interface InterpreterDirective <: Node {
+    type: "InterpreterDirective";
+    value: string;
+}
+```
+
+The `value` of an intepreter directive is the content without `#!` prefix, e.g. `/usr/bin/env node`.
+
+[hashbang grammar]: https://github.com/tc39/proposal-hashbang


RReverser:
Hmm I'm not sure about this one. Hash bang is more of a comment, really, and has no associated JavaScript semantics.

Acorn supported it under a separate option for couple of years now, and used to just ignore it as any other comment.

To me, it seems like that's what ESTree should do as well here.

nzakas:
I think this comes down to whether or not we want to treat this as a comment. If it's a comment, then it would make sense to leave out of ESTree; if it's a directive, then it would make sense to include. I can see an argument either way.

RReverser:
@nzakas This is probably a good generic question for #219, but in general the way I see it is that anything that doesn't have associated JavaScript semantics, doesn't matter for us.

For example, comments like `//eslint-ignore` or `//@ts-check` could be argued to be directives as well, but in the context of ECMAScript, and, correspondingly, ESTree, they don't matter since they have semantics only within their respective runtime environments.

Hashbang is very similar in this regard - it doesn't matter in context of ECMAScript / ESTree, but has semantics in its respective runtime environment (Unix).

mysticatea:
As per the spec, shebang is a comment:

> Hashbang Comments are location-sensitive and like other types of comments are discarded from the stream of input elements for the syntactic grammar.
> ```
> HashbangComment::
>     #! SingleLineCommentCharsopt_opt
> ```
> https://tc39.es/proposal-hashbang/out.html#prod-HashbangComment

Therefore, ESTree's AST should not contain it as same as other comments.

RReverser:
Thanks @mysticatea! @nzakas @JLHwung - do you agree given the new information above, should we close this?

nzakas:
I'm fine with treating as a comment.

JLHwung:
Closing this as hashbang is considered a comment and thus should not be included in estree.

PR #229, opened 2020-07-03
add slice annotation

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/slice-notation/experimental/slice-notation.md)

This PR adds stage-1 [Slice notation] support.

[Slice notation]: https://github.com/tc39/proposal-slice-notation

Note for reviewer: A minimal spec text is recently drafted: some README examples may not reflect on these changes.

### DIFF
--- a/experimental/slice-notation.md
+++ b/experimental/slice-notation.md
@@ -1,0 +1,16 @@
+# [Slice notation][proposal-slice-notation]
+
+## SliceExpression
+
+```js
+interface SliceExpression <: ChainElement, Expression {
+    type: "SliceExpression";
+    object: Expression;
+    lower: Expression | null;
+    upper: Expression | null;
+}
+```
+
+A slice expression, e.g. `arr[0:1]`.
+
+[proposal-slice-notation]: https://github.com/tc39/proposal-slice-notation


experimental/slice-notation.md
```diff
@@ -0,0 +1,16 @@
+# [Slice notation][proposal-slice-notation]
+
+## SliceExpression
+
+```js
+interface SliceExpression <: ChainElement, Expression {
+    type: "SliceExpression";
+    object: Expression;
+    lower: Expression | null;
```
RReverser, line 9: Nitpick (which you're free to ignore for now): I'd be more explicit in naming and call it `lowerBound` and `upperBound`.

JLHwung: I don't have strong opinion on naming. As this is a new AST type, I am open to name changes as long as we can reach consensus.

This feature is not yet implemented in Babel so we have chances to respond to any AST breaking changes. Ideally Babel should follow ESTree as much as possible so we don't cause downstream pains when and proposal advances to stage 4.

RReverser: Yep, sounds good, as I said, this is just a nitpick that can be ignored if you wish :)

nzakas: I think maybe, per previous conversations, we should leave nitpicks out of the review of experimental features? If we're saying that we're only doing a cursory glance at stage 1 features, I'd like if we can agree to streamline by just pointing out any big issues.

Of course, if we want to say that we'll do more thorough reviews of experimental features, that's fine, I'd just like us to be consistent one way or the other.

RReverser: Nah, as I said above, "feel free to ignore for now". That is, it's more of a general comment that doesn't block merging in any way, just to point something out for the future, but if you think those are confusing, I can refrain from leaving them :)

RReverser: That is, I think this is good to merge either way, nitpicks in my mind is mostly meant as a general... well, a nitpick :D

JLHwung: @nzakas WDYT of `upper` vs `upperBound`? If you two prefer `upperBound` over `upper`, I can revise the draft and follow the majority here. I expect the AST shape does not change ... due to only naming conventions, because it will pose downstream breaking changes if Babel implements `upper` and the ESC later recommends `upperBound`.

That said, if the proposal changes significantly in later stages, i.e. if `lower` is completely removed and a new `stride` option is added. Then we _have to_ make breaking changes to the AST spec. Otherwise I expect we can reach naming consensus.

RReverser: Oh no, let's definitely not go the full consensus process for experimental stuff. If it's not the sort of change you immediately agree with, let's just merge it as-is for now :)

nzakas: @JLHwung I personally think that "bound" is implicit in the usage, but have no objections to using `upperBound` and `lowerBound`.

@RReverser I think the problem with nitpicks, even when you say "don't block merging" is that you're sending two messages: 1) that you have something you want to mention and 2) it's not important enough to prevent merging. If both of those conditions are true, I think we can streamline by either holding those comments until the syntax has reached stage 3 or maybe instead of "nitpick", flag as something like "Questions to consider at stage 3:" to make it clear what the intention is. Again, I'm just trying to better define the process for things so we can document it and understand expectations.

I'd still like a chance to review experimental syntax before it's merged, as I think it's only fair to give everyone time to take a look. However, maybe we could impose a time limit to say that a pull request can be merged even without everyone approving if it's been longer than 14 days since the PR was opened? That way someone disappearing doesn't block anything.

nzakas:
It seemed like we all agreed to merge as-is right now, so I'm going to go ahead and merge. We can follow up on other process-related questions elsewhere.

PR #231, opened 2020-07-13
add class-static-block

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/class-static-block/experimental/class-static-block.md)

This PR adds [class static initialization blocks] support. The proposal is seeking stage 2 advancement in July meeting.

[class static initialization blocks]: https://github.com/tc39/proposal-class-static-block

### DIFF
--- a/experimental/class-static-block.md
+++ b/experimental/class-static-block.md
@@ -1,0 +1,22 @@
+# [class static initialization blocks][proposal-class-static-block]
+
+## ClassBody
+
+```js
+extend interface ClassBody {
+    body: [ MethodDefinition | PropertyDefinition | StaticBlock ];
+}
+```
+- `body` has at most one `StaticBlock`.
+
+## StaticBlock
+
+```js
+interface StaticBlock <: BlockStatement {
+    type: "StaticBlock"
+}
+```
+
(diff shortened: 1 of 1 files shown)


nzakas:
@RReverser thoughts?

experimental/class-static-block.md
```diff
@@ -0,0 +1,22 @@
+# [class static initialization blocks][proposal-class-static-block]
+
+## ClassBody
+
+```js
+extend interface ClassBody {
+    body: [ MethodDefinition | PropertyDefinition | StaticBlock ];
+}
+```
+- `body` has at most one `StaticBlock`.
```
RReverser, line 10: I'm somewhat tempted to suggest moving it out in a separate property in that case... but I can see the argument for preserved ordering being useful so idk. I guess we can leave it for later.

RReverser (approved):

experimental/class-static-block.md
JLHwung: Yeah I agree. The static block is sort-of like the `constructor`, in a way that it has special semantics but can mix with other class elements in any order. To align to current situation that `constructor` is a `MethodDefinition` in `body`, I go with extending current body.

PR #234, opened 2020-08-09
added await.ops

anikethsaha:
Added [await.ops](https://github.com/tc39/proposal-await.ops) (`stage-0`) to `experiment`

**Question to reviewers** : 
Should it be just as addition to existing `AwaitExpression`

```js
interface AwaitExpression <: Expression {
    operation: "all" | "any" | "race" | "allSettled" | undefined
}
```
or the full one that is currently being implemented.

### DIFF
--- a/experimental/await-ops.md
+++ b/experimental/await-ops.md
@@ -3,10 +3,8 @@
 ## AwaitExpression
 
 ```js
-interface AwaitExpression <: Expression {
-    type: "AwaitExpression";
-    argument: Expression;
-    operation: Identifier | null
+extend interface AwaitExpression {
+    operation: Identifier | null // operation.name should be one of the following : 'all', 'race', 'allSettled' and 'any'
 }
 ```
 
--- a/experimental/await-ops.md
+++ b/experimental/await-ops.md
@@ -6,7 +6,7 @@
 interface AwaitExpression <: Expression {
     type: "AwaitExpression";
     argument: Expression;
-    operation: "all" | "any" | "race" | "allSettled" | undefined
+    operation: Identifier | null
 }
 ```
 
--- a/experimental/await-ops.md
+++ b/experimental/await-ops.md
@@ -1,0 +1,22 @@
+# [Await.ops](https://github.com/tc39/proposal-await.ops)
+
+## AwaitExpression
+
+```js
+interface AwaitExpression <: Expression {
+    type: "AwaitExpression";
+    argument: Expression;
+    operation: "all" | "any" | "race" | "allSettled" | undefined
+}
+```
+
+eg :
+
+```js
+async function fn() {
+    await [];
+    await.all [];
+    await.race [];
+    await.allSettled [];
+}
+```


experimental/await-ops.md
```diff
@@ -0,0 +1,22 @@
+# [Await.ops](https://github.com/tc39/proposal-await.ops)
+
+## AwaitExpression
+
+```js
+interface AwaitExpression <: Expression {
+    type: "AwaitExpression";
+    argument: Expression;
+    operation: "all" | "any" | "race" | "allSettled" | undefined
```
j-f1, line 9: To allow easy extensibility, should this be an `Identifier | null` so future additions won’t have to change the AST or does it make sense to strictly limit the allowed operations?

anikethsaha: In the typescript implementation of this [(repl)](https://www.staging-typescript.org/play?ts=4.0.0-pr-39224-4#code/IYZwngdgxgBAZgV2gFwJYHsIwB4AoCUMA3gFAzkzADuwqyAdMADZMwAMZF1tDATsFACmMANoAFXugC2qEIPq9BIdEwBug3AEZ8AXU7ludRiwDKg5MiaCAJqL0BfIA), it does throw an error so I think we need it to be strict. 

Though I don't have strong opinion on either

JLHwung: It seems to me that current spec disallows whitespace in `await.all`, which TypeScript implementation currently allow
```js
await
// some comment
   .all [Promise.resolve(1)]
```
then an enum looks good to me. But I believe the spec should allow whitespaces. I have submitted https://github.com/tc39/proposal-await.ops/issues/12.

If whitespaces are allowed, an `Identifier | null` will offer position informations for tools. We had similar discussion about that on `MetaProperty` ([thread](https://github.com/estree/estree/pull/32#issuecomment-77819820)).

nzakas: Also agree that an enum is most appropriate here as anything other than these values would be a syntax error.

JLHwung: Sorry for the wordings before. I lean to a dedicated Identifier for tracking positional information of `all` _if_ whitespaces are allowed: https://github.com/tc39/proposal-await.ops/pull/13

RReverser: +1 for enum. But, either way, let's remember this is an experimental PR ;)

JLHwung, line 9:
```suggestion
    operation: Identifier | null
```
Whitespaces are now allowed in `await.ops`. I suggest we align with what we did in [MetaProperty](https://github.com/estree/estree/blob/master/es2015.md#metaproperty).

anikethsaha: Updated the PR with `Indentifier`

experimental/await-ops.md
```diff
@@ -0,0 +1,22 @@
+# [Await.ops](https://github.com/tc39/proposal-await.ops)
+
+## AwaitExpression
+
+```js
+interface AwaitExpression <: Expression {
+    type: "AwaitExpression";
+    argument: Expression;
```
JLHwung, line 8:
```suggestion
extend interface AwaitExpression {
```
An editorial update. All experimental proposals are filed as a diff to current spec.

experimental/await-ops.md
```diff
@@ -0,0 +1,22 @@
...
+    operation: Identifier | null
+}
+```
```
JLHwung, line 11: As now the operation is an `Identifier`, can you add a comment that the `operation.name` must be one of `"all" | "allSettled" | "race" | "any"`?

experimental/await-ops.md
anikethsaha: done thanks,

experimental/await-ops.md
anikethsaha: I added an inline comment , let me know if that's okay.

PR #237, opened 2020-09-09
add class-access-expressions

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/class-access-expressions/experimental/class-access-expressions.md)

This PR adds [Class Property Access Expressions] support. The proposal is seeking stage 2 advancement in September meeting.

[Class Property Access Expressions]: https://github.com/tc39/proposal-class-access-expressions

### DIFF
--- a/experimental/class-access-expressions.md
+++ b/experimental/class-access-expressions.md
@@ -1,0 +1,23 @@
+# [class property access expressions][proposal-class-access-expressions]
+
+## ClassObject
+
+```js
+interface ClassObject <: Node {
+    type: "ClassObject";
+}
+```
+
+A `class` pseudo-expression, e.g. `class.foo`.
+
+## MemberExpression
+
+```js
+extend interface MemberExpression {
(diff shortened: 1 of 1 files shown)


experimental/class-access-expressions.md
```diff
@@ -0,0 +1,23 @@
+# [class property access expressions][proposal-class-access-expressions]
+
+## ClassObject
+
+```js
+interface ClassObject <: Node {
+    type: "ClassObject";
```
JLHwung, line 7: The `ClassObject` is meant to be aligned with `Super`. Unfortunately we can not name it `Class` because `Class` has been used for class declarations / expressions, which is more common.

I pick `ClassObject` from the spec text. Any naming suggestion is welcome.

nzakas (approved):
Naming is hard and I can’t think of a better one, so makes sense for now.

RReverser (approved):

Issue #240, opened 2020-12-30
Supporting Stage 3 Features


bradzacher:
I just wanted to start a discussion around the philosophy of this project to only officially support Stage 4 features.

I fully understand the reason behind only supporting Stage 4 features:
As listed in the [TC39 process document](https://tc39.es/process-document/), a Stage 3 feature may experience "Limited Post-Acceptance Changes Expected: only those deemed critical based on implementation experience".
If one of these critical changes causes a syntax change, then it means a breaking change in ESTree (which has no versioning, so handling that will be very `undefined`).

However - I would question the likelihood that a Stage 3 feature has a change that requires a change in the AST. By the time a feature reaches Stage 3:
- it's been through many months of iteration and review
- there's a general expectation that there is already a spec-complaint implementation of the feature

Which means (from my understanding) that it's very unlikely that the feature would undergo syntax changes - it's more likely that the feature would undergo changes in its implementation details.

The reason that I bring this up is because, by-and-large, the JS tooling ecosystem is converging on Stage 3 as being the minimum stable stage for implementation.

By not supporting Stage 3, it means tools like eslint cannot support it by default - leading to a non-trivial number of users that opt-in to using a 3rd-party parser which supports the newer syntax.

This also leads the ESTree ecosystem into a difficult position - how do you support Stage 3 features?
Generally this means that "the biggest project" moves to implement it, and everyone follows suit to match their AST for compatibility.
But this leads projects into issues if this AST does not match the final implementation.

Two recent examples: optional chaining and class features. For both of these the ecosystem converged on babel's AST.
- For optional chaining - an entirely different AST structure was specced for ESTree - which was a large breaking change for implementors of babel's AST.
- For class features - it's not yet a "stable" ESTree AST as it's stage 3, but the [experimental AST that's currently merged](https://github.com/estree/estree/blob/390a050fd2d6a7b69688f1434e0929d7c7455d05/experimental/class-features.md) has different naming to babel's AST (`ClassProperty -> PropertyDefinition`) and also some minor structure changes (`PrivateName -> PrivateIdentifier`).

These breakage causes *huge* pains both for maintainers and for users:
- parser maintainers have to scramble to release major version bumps to support the new AST to help the ecosystem move forward.
- eslint plugin maintainers have to make a choice about when / how to support the new AST
  - this can be non-trivial when 
- users have to navigate the fragmented ecosystem to figure out which module works with what version and how to piece it all together.
[...]

PR #241, opened 2021-02-12
Introduce "stage3" vs. "experimental" distinction

tchetwin:
Following https://github.com/estree/estree/issues/240#issuecomment-778168995, this promotes the currently "experimental" class features AST specification in to the root.

### DIFF
--- a/experimental/README.md
+++ b/experimental/README.md
@@ -1,3 +1,5 @@
-These documents specify extensions to the core ESTree AST types to support ES proposals that are at least [stage 0 or above](https://github.com/tc39/ecma262).
+These documents specify extensions to the core ESTree AST types to support ES
+proposals that are in [stages 0, 1 & 2](https://github.com/tc39/proposals).
 
-**NOTE:** These documents may change at anytime and **should not** be seen as stable.
+**NOTE:** These documents may change at anytime and **should not** be seen as
+stable.
--- a/stage3/README.md
+++ b/stage3/README.md
@@ -1,0 +1,6 @@
+These documents specify extensions to the core ESTree AST types to support ES
+proposals that are at [stage 3](https://github.com/tc39/proposals#stage-3).
+
+**NOTE:** Concensus has been reached on the AST shapes and they can largerly be
+considered stable for implementers. The shape will only change if the spec
+undergoes significant changes from stage 3 to stage 4.


nzakas:
This isn’t the correct way to promote nodes. All files have an ES version number, so this should probably be es2022.md. Please see es2021.md as an example for how to set up the file.

JLHwung (commented):
I think we should create es2022 only if class features reaches stage 4 officially after es2021 is standardized. Before that it is experimental by definition.

tchetwin:
I'm somewhat inclined to agree, picking `es2021`/`es2022`/`es20**` is forecasting. Perhaps just `esnext.md`?

RReverser:
@tchetwin That's essentially what individual features in experimental folder represent - any ESNext features that are not yet finalized in any ES version.

Another doc for the same purpose seems unnecessary.

bradzacher:
The problem is that "experimental" is unclear. Does it mean the AST is experimental, the feature is experimental, or both? 

Some clarity should be added, esp if the intention is to allow stage 3 features to be considered stable (see #240). 

If a new file called "esnext" isn't desired, then maybe a separate folder called such is better? Something to mark an AST spec as "ready for use but feature is stage 3, not stage 4" so it's separate from the "still experimental and subject to change" ast specs.

nzakas:
I do think it's useful to have a distinction between "experimental" (stage 0-2) and "baked" (stage 3-4). We know that stage 3 features are effectively done and so if we are happy with the AST format for those, it would be nice to have some way to flag those as separate from the < stage 3 features.

If we don't want an `esnext.md`, then perhaps a `stage3` folder could serve the same purpose?

JLHwung:
> a stage3 folder could serve the same purpose?

That works for me. On those PRs moving from `experimental` to `stage3`, we should reach consensus on the AST shapes and it can be considered stable for implementers, with a condition that the AST will only change unless the spec has undergone significant changes from stage 3 to stage 4. By doing that we are effectively advancing AST shape consensus from stage 4 to stage 3 proposals.

tchetwin:
I am *very* happy to see us draw a distinction between `experimental` (stages <3) and `stage3`. I think sending this signal will make it a lot easier for tooling to confidently make progress on stage3 features and themselves provide feedback in the stage advancement process.

RReverser:
Separate `stage3` sounds reasonable to me. I wonder if it should be a single doc instead of a folder perhaps - since it's almost same as stable `es*.md` versions?

tchetwin:
> then perhaps a `stage3` folder could serve the same purpose?

> I wonder if it should be a single doc instead of a folder perhaps

Could go either way, `stage3.md` or `stage3/class-features.md`

A single file might "force" reconciliation of proposals that touch similar parts of ASTs. Imagine if [class-static-block](https://github.com/estree/estree/blob/master/experimental/class-static-block.md) was also stage 3, it also changes `ClassBody`, having it in a different file might be confusing. This reconciliation would be similar to how the `acorn-stage3` collects various underlying plugins under one umbrella.

Counter to that: if one of two interlinked proposals makes it from stage3 -> stage4 before the other, they then need to be untangled again. Some proposals (like class-features) have spent a long time in stage3 and have been overtaken by other proposals.

stage3/class-features.md
```diff
@@ -1,7 +1,6 @@
+# Classes
```
JLHwung, line 1: Can you separate this one to another PR? So this PR focuses on creating the `stage3` thing and in another PR we seek consensus on the specific class features.

tchetwin: I can do, these aren't really material changes to the class features themselves though, just a massaging of the format to more closely resemble other files.

tchetwin: I guess you're implicitly advocating for the `stage3/*.md` model as otherwise there is no other change to be made.

tchetwin:
On @JLHwung's recommendation I have removed the class-features specifics so this PR now only introduced the concept of a distinction between "stage3" and "experimental". Class features promotion will follow in another PR.

JLHwung (approved):

nzakas:
I think it's a good idea to keep separate files in the `stage3` folder like we do in `experimental` because it will limit the number of time we need to copy-paste into a single file. :+1:

nzakas (approved):

RReverser (approved):

PR #250, opened 2021-04-19
Add ES2022

JLHwung:
Class features just reach stage 4: https://github.com/tc39/proposals/blob/master/finished-proposals.md


tchetwin:
I came here to find this 😁

PR #251, opened 2021-06-18
add pattern matching

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/pattern-matching/experimental/pattern-matching.md)

See also [AST examples](ast-examples) for this PR.

This PR adds support to the [Pattern matching] proposal. Besides the `match () {}` expression, the proposal introduces the concept of `MatchPattern`, which has subtle differences with current destructuring `Pattern`. I copied current `Pattern` design and infix the Node type with `Match`. I am open to changes of AST structures or renaming of AST node shapes.

[Pattern matching]: https://github.com/tc39/proposal-pattern-matching/
[ast-examples]: https://gist.github.com/JLHwung/d2b64364dc63757282cef182367b4e84

### DIFF
--- a/experimental/pattern-matching.md
+++ b/experimental/pattern-matching.md
@@ -1,0 +1,119 @@
+# [Pattern Matching](proposal-pattern-matching)
+
+See also [AST examples](ast-examples) for this proposal.
+
+## MatchExpression
+
+```js
+interface MatchExpression <: Expression {
+    type: "MatchExpression";
+    discriminant: Expression;
+    id: Pattern | null;
+    clause: [ MatchClause ]
+}
+```
+
+## WhenClause
+```js
+interface MatchClause <: Node {
+    type: "MatchClause";
+    test: MatchPattern | null;
+    guard: Expression | null;
+    id: Pattern | null;
+    consequent: BlockStatement
+}
+```
+
+If `test` is `null`, `id` must be `null`.
+
+## MatchPattern
+```js
+interface MatchPattern <: Node {}
+```
+
+## ArrayMatchPattern
+```js
+interface ArrayMatchPattern <: MatchPattern {
+    type: "ArrayMatchPattern";
+    elements: [ MatchPattern ];
+}
+```
+
+## ObjectMatchPattern
+```js
+interface AssignmentMatchProperty <: Property {
+    type: "Property"; // inherited
+    value: MatchPattern;
+    kind: "init";
+    method: false;
+}
+
+interface ObjectMatchPattern <: MatchPattern {
+    type: "ObjectMatchPattern";
+    properties: [ AssignmentMatchProperty ];
+}
+```
+
+## RestMatchElement
+```js
+interface RestMatchElement <: MatchPattern {
+    type: "RestMatchElement";
+    argument: MatchPattern;
+}
+```
+
+## BinaryMatchPattern
+```js
+interface BinaryMatchPattern <: MatchPattern {
+    type: "BinaryPattern";
+    operator: "and" | "or" | "with";
+    left: MatchPattern;
+    right: MatchPattern;
+}
+```
+
+## AsMatchPattern
+```js
+interface AsMatchPattern <: MatchPattern {
+    type: "AsMatchPattern";
+    test: MatchPattern;
+    id: Pattern;
+}
+```
+
+## ExpressionMatchPattern
+```js
+interface ExpressionMatchPattern <: MatchPattern {
+    type: "ExpressionMatchPattern";
+    expression: Expression;
+}
+```
+
+## NullMatchPattern
+```js
+interface NullMatchPattern <: MatchPattern {
+    type: "NullMatchPattern";
+}
+```
+
+# Literal
+
+```js
+extend interface Literal <: Expression, MatchPattern {}
+```
+
+# Identifier
+```js
+extend interface Identifier <: Expression, Pattern, MatchPattern {}
+```
+
+### UnaryExpression
+
+```js
+extend interface UnaryExpression <: Expression, MatchPattern {}
+```
+
+If a UnaryExpression is a MatchPattern, its `operator` must be `"-"`, its `prefix` must be `true`, and its `argument` must be either a Literal with numeric value, or an identifier named `"Infinity"`.
+
+[proposal-pattern-matching]: https://tc39.es/proposal-pattern-matching/
+[ast-examples]: https://gist.github.com/JLHwung/d2b64364dc63757282cef182367b4e84


experimental/pattern-matching.md
```diff
@@ -0,0 +1,119 @@
...
+    method: false;
+}
+
+interface ObjectMatchPattern <: MatchPattern {
+    type: "ObjectMatchPattern";
+    properties: [ AssignmentMatchProperty ];
+}
+```
+
+## RestMatchElement
```
JLHwung, line 57: The `ArrayMatchPattern`, `ObjectMatchPattern` and `RestMatchElement` are essentially `ArrayPattern`, `ObjectPattern` and `RestElement` with `Pattern` replaced by `MatchPattern`.

experimental/pattern-matching.md
```diff
@@ -0,0 +1,119 @@
...
+    left: MatchPattern;
+    right: MatchPattern;
+}
+```
+
+## AsMatchPattern
+```js
+interface AsMatchPattern <: MatchPattern {
+    type: "AsMatchPattern";
+    test: MatchPattern;
```
JLHwung, line 79: The `AsMatchPattern` is separated from `BinaryMatchPattern` because the `right` (renamed as `id`) accepts a `Pattern`, not `MatchPattern`.

experimental/pattern-matching.md
```diff
@@ -0,0 +1,119 @@
...
+    type: "RestMatchElement";
+    argument: MatchPattern;
+}
+```
+
+## BinaryMatchPattern
+```js
+interface BinaryMatchPattern <: MatchPattern {
+    type: "BinaryPattern";
+    operator: "and" | "or" | "with";
```
JLHwung, line 69: The champions have not decided the combinator token of `"and"` and `"or"`: https://github.com/tc39/proposal-pattern-matching/issues/179. We could change to `&` and `|` here if they go with these sigils.

experimental/pattern-matching.md
```diff
@@ -0,0 +1,119 @@
...
+}
+```
+
+## ExpressionMatchPattern
+```js
+interface ExpressionMatchPattern <: MatchPattern {
+    type: "ExpressionMatchPattern";
+    expression: Expression;
+}
+```
```
ljharb, line 90: not sure if this is covered or not; the syntax allows `^Identifier`, `^(Expression)`, but also things of the form `^A['B'].C()` or `^A?['B']?.C?.()`.

experimental/pattern-matching.md
```diff
@@ -0,0 +1,119 @@
...
+extend interface Identifier <: Expression, Pattern, MatchPattern {}
+```
+
+### UnaryExpression
+
+```js
+extend interface UnaryExpression <: Expression, MatchPattern {}
+```
+
+If a UnaryExpression is a MatchPattern, its `operator` must be `"-"`, its `prefix` must be `true`, and its `argument` must be either a Literal with numeric value, or an identifier named `"Infinity"`.
```
ljharb, line 116: as a pattern, this also allows `undefined` (even if the identifier resolves to something that's not the `undefined` value), but also, all number and bigint literals (with +, -, or no sign at all), `null`, booleans, etc.

Also, template literals haven't been thought out here - i filed https://github.com/tc39/proposal-pattern-matching/issues/205 to track

experimental/pattern-matching.md
bradzacher: does it matter if there are parentheses or not?

ljharb: Yes; any expression can be in parens, but only a specific limited set of forms can omit them.

experimental/pattern-matching.md
JLHwung: Yeah, `undefined` is covered in the `Identifier` case

```
extend interface Identifier <: Expression, Pattern, MatchPattern {}
```

The `UnaryExpression` case is addressed to things like `-0`, `-1`, `-1n`, since they are an UnaryExpression from the AST perspective.

The template literals in ESTree has a specific type `TemplateLiteral`, in this PR the `TemplateLiteral` is not mentioned so it is not a `MatchPattern`. But yeah it is a good question whether it should be allowed. I will update this PR if necessary.

experimental/pattern-matching.md
```diff
@@ -0,0 +1,119 @@
...
+See also [AST examples](ast-examples) for this proposal.
+
+## MatchExpression
+
+```js
+interface MatchExpression <: Expression {
+    type: "MatchExpression";
+    discriminant: Expression;
+    id: Pattern | null;
+    clause: [ MatchClause ]
```
nzakas, line 12:
```suggestion
    clauses: [ MatchClause ]
```

Should this be plural give that it’s an array?

experimental/pattern-matching.md
```diff
@@ -0,0 +1,119 @@
...
+    test: MatchPattern | null;
+    guard: Expression | null;
+    id: Pattern | null;
+    consequent: BlockStatement
+}
+```
+
+If `test` is `null`, `id` must be `null`.
+
+## MatchPattern
```
nzakas, line 29: Could we add a mention here as to why we don’t just use Pattern? (You mentioned in the PR itself but seems important to include here as well.)

nzakas (approved):
Overall looks good. I haven’t had the time to dig in and explore the proposal at any depth but hope too soon.

experimental/pattern-matching.md
JLHwung: > the syntax allows ^Identifier, ^(Expression), but also things of the form ^A['B'].C() or ^A?['B']?.C?.().

Both `^Identifier` and `^(Expression)` are an ExpressionMatchPattern whose `expression` is an `Identifier` or an `Expression`. ESTree does not track whether parentheses are required to form a valid syntax. Instead a parser will throw when parentheses are required and a generator will add parentheses based on the AST structure.

experimental/pattern-matching.md
JLHwung: Sure, addressed in d67b89a.

experimental/pattern-matching.md
```diff
@@ -0,0 +1,126 @@
...
+interface RestMatchElement <: MatchPattern {
+    type: "RestMatchElement";
+    argument: MatchPattern;
+}
+```
+
+## BinaryMatchPattern
+```js
+interface BinaryMatchPattern <: MatchPattern {
+    type: "BinaryPattern";
```
fedeci, line 75: Should this be `BinaryMatchPattern` for consistency?

JLHwung: Good catch!

experimental/pattern-matching.md
```diff
@@ -0,0 +1,126 @@
...
+interface AssignmentMatchProperty <: Property {
+    type: "Property"; // inherited
+    value: MatchPattern;
+    kind: "init";
+    method: false;
+}
+
+interface ObjectMatchPattern <: MatchPattern {
+    type: "ObjectMatchPattern";
+    properties: [ AssignmentMatchProperty ];
```
fedeci, line 60:
```suggestion
    properties: [ AssignmentMatchProperty | Property ];
```

JLHwung: Can you elaborate? `AssignmentMatchProperty` inherits `Property` but its value is limited to a `MatchPattern`. A general object property can not fit here, e.g.

```js
match (x) {
  when ({ a: () => {} }) {}
}
```
is invalid because `() => {}` is not a match pattern.

fedeci: Sure, sorry. If `ObjectMatchPattern` only accepts `AssignmentMatchProperty` as its properties, that means that even
`{ status: 200 | 201 }` should have `"kind": "init"`. But `status` should not create a new binding, right?

JLHwung: Yes. In ESTree a pattern element in an ObjectPattern always has `"kind": "init"`.  The ObjectMatchPattern is aligned with ObjectPattern. 

> `status` should not create a new binding

Nope. A match pattern creates bindings (if matched). See the code examples on the proposal:

```js
match (res) {
  if (isEmpty(res)) { … }
  when ({ data: [page] }) { … }
  when ({ data: [frontPage, ...pages] }) { … }
  else { … }
}
```

> The first when clause matches if data has exactly one element, and binds that element to page for the right-hand side.

fedeci: My doubts concern the examples provided above where `Property` is used instead of `AssignmentMatchProperty`.

JLHwung: I am not sure I can follow. Can you elaborate why `properties` of an `ObjectMatchPattern` should accept general `Property`?

The `type` of `AssignmentMatchProperty` is `"Property"`. It aligns to how ESTree models an object pattern: `a: 1` in `({ a: 1 } = {})` is a `property`. See also https://github.com/estree/estree/blob/master/es2015.md#objectpattern

fedeci: It's clear now, I though that `AssignmentMatchProperty` would have been a new node🤦

experimental/pattern-matching-examples.md
```diff
@@ -0,0 +1,373 @@
...
+                        "key": { "type": "Identifier", "name": "y" }
+                    }]
+                }
+            }]
+        }
+    }]
+}
+```
+
+## 4 "if" Match Clause and Binary Match Pattern
```
RReverser, line 167: Nit: missing dots here and below in headings.

experimental/pattern-matching-examples.md
```diff
@@ -0,0 +1,373 @@
+# AST examples of [Pattern Matching ESTree Proposal](proposal-pattern-matching-estree)
```
RReverser, line 1: Why not keep this as a section in the main `.md`?

JLHwung: Oh the examples are not supposed to be part of the spec. They serves as an explainer only.

RReverser:
It's rather hard to understand which of the conversations here are still unresolved. Is there anything left?

ljharb:
It's worth noting that the syntax is still in flux; for example, we'll probably go with `and`/`or` for combining patterns, and it's possible we'll drop `as` entirely.

nzakas:
@JLHwung can you mark conversations as resolved so we can see what’s left to discuss here?

@ljharb this is being submitted to the experimental folder, so we aren’t committing forever, just to have a format we can work with at least in the short term

JLHwung:
> it's possible we'll drop as entirely.

Yeah the AST spec tries to cover all the mentioned syntax. If we drop some features when the proposal gets advanced, we can revise the AST accordingly. The experimental folder is deemed unstable and we are free to make breaking changes.

I have resolved current review issues. Let me know if there's more. I am on vacation and will come back to work on it next week.

PR #255, opened 2021-09-09
add grouped-and-auto-accessors

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/grouped-and-auto-accessors/experimental/grouped-and-auto-accessors.md)

This PR adds support to the [Grouped Accessors and Auto-Accessors] proposal. 

[Grouped Accessors and Auto-Accessors]: https://github.com/tc39/proposal-grouped-and-auto-accessors

### DIFF
--- a/experimental/grouped-and-auto-accessors.md
+++ b/experimental/grouped-and-auto-accessors.md
@@ -1,0 +1,85 @@
+# [Grouped Accessors and Auto-Accessors][proposal-grouped-and-auto-accessors]
+
+## ClassBody
+
+```js
+extend interface ClassBody {
+    body: [ MethodDefinition | PropertyDefinition | StaticBlock | ClassAccessor ];
+}
+```
+
+## ObjectExpression
+
+```js
+extend interface ObjectExpression {
+    properties: [ Property | ObjectAccessor ];
+}
+```
+
+## ClassAccessor
+
+```js
+interface ClassAccessor <: PropertyDefinition {
+    type: "ClassAccessor";
+    get: GetAccessorStub | GetAccessorMethod | null;
+    set: SetAccessorStub | SetAccessorMethod | null;
+}
+```
+
+If `set.private` is `true`, `key` must be an `Identifier` and `computed` must be `false`.
+
+## ObjectAccessor
+
+```js
+interface ObjectAccessor <: Property {
+    type: "ObjectAccessor";
+    get: GetAccessorStub | GetAccessorMethod | null;
+    set: SetAccessorStub | SetAccessorMethod | null;
+}
+```
+
+### GetAccessorStub
+
+```js
+interface GetAccessorStub <: Node {
+    type: "GetAccessorStub";
+}
+```
+
+A get accessor stub is a simplified version of grouped accessor. E.g. `class C { accessor c { get; } }`.
+
+### GetAccessorMethod
+
+```js
+interface GetAccessorMethod <: Node {
+    type: "GetAccessorMethod";
+    params: [];
+    body: FunctionBody;
+}
+```
+
+### SetAccessorStub
+
+```js
+interface SetAccessorStub <: Node {
+    type: "SetAccessorStub";
+    private: boolean;
+}
+```
+
+A set accessor stub is a simplified version of grouped accessor. E.g. `class C { accessor c { set; } }`.
+
+### SetAccessorMethod
+
+```js
+interface SetAccessorMethod <: Node {
+    type: "SetAccessorMethod";
+    private: boolean;
+    params: [ Pattern ];
+    body: FunctionBody;
+}
+```
+
+The `params` of a `SetAccessorMethod` must include only one element, which must not be a `RestElement`.
+
+[proposal-grouped-and-auto-accessors]: https://github.com/tc39/proposal-grouped-and-auto-accessors


experimental/grouped-and-auto-accessors.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+A get accessor stub is a simplified version of grouped accessor. E.g. `class C { accessor c { get; } }`.
+
+### GetAccessorMethod
+
+```js
+interface GetAccessorMethod <: Node {
+    type: "GetAccessorMethod";
+    params: [];
```
JLHwung, line 56: The `params` must be an empty array. It is to provide compatibility for general function AST visitors: A function always has `params` property.

experimental/grouped-and-auto-accessors.md
```diff
@@ -0,0 +1,85 @@
...
+
+```js
+interface ObjectAccessor <: Property {
+    type: "ObjectAccessor";
+    get: GetAccessorStub | GetAccessorMethod | null;
+    set: SetAccessorStub | SetAccessorMethod | null;
+}
+```
+
+### GetAccessorStub
```
nzakas, line 41: Why use “stub” instead of using the proposal language, such as GetAutoAccessor?

experimental/grouped-and-auto-accessors.md
```diff
@@ -0,0 +1,85 @@
...
+
+```js
+interface ClassAccessor <: PropertyDefinition {
+    type: "ClassAccessor";
+    get: GetAccessorStub | GetAccessorMethod | null;
+    set: SetAccessorStub | SetAccessorMethod | null;
+}
+```
+
+If `set.private` is `true`, `key` must be an `Identifier` and `computed` must be `false`.
```
nzakas, line 29: Maybe this is a bit clearer?

```suggestion
If `set.private` is `true`, `ClassAccessor#key` must be an `Identifier` and `ClassAccessor#computed` must be `false`.
```

experimental/grouped-and-auto-accessors.md
JLHwung: The "stub" comes from the [proposed syntax section](https://github.com/tc39/proposal-grouped-and-auto-accessors#proposed-syntax).

According to the spec, an auto-accessor is a simplified version of grouped accessor, e.g. `accessor x = { get; set; }`, where the accessor stub refers to the `set` and `get` shorthand.

nzakas: Ah ok.

nzakas (approved):

experimental/grouped-and-auto-accessors.md
```diff
@@ -0,0 +1,85 @@
...
+}
+```
+
+## ClassAccessor
+
+```js
+interface ClassAccessor <: PropertyDefinition {
+    type: "ClassAccessor";
+    get: GetAccessorStub | GetAccessorMethod | null;
+    set: SetAccessorStub | SetAccessorMethod | null;
```
pzuraq, line 25: `ClassAccessor` should also have a `value` field which is similar to a class field's value/initializer:

```js
accessor foo { get; set; } = 123;
```

pzuraq: Ah, I see, it inherits these properties. I think this makes sense.

pzuraq (approved):

experimental/grouped-and-auto-accessors.md
```diff
@@ -0,0 +1,85 @@
...
+### SetAccessorStub
+
+```js
+interface SetAccessorStub <: Node {
+    type: "SetAccessorStub";
+    private: boolean;
+}
+```
+
+A set accessor stub is a simplified version of grouped accessor. E.g. `class C { accessor c { set; } }`.
```
JLHwung, line 70:
```suggestion
A set accessor stub is used in auto-accessor. E.g. `class C { accessor c { set; } }`.
```

experimental/grouped-and-auto-accessors.md
```diff
@@ -0,0 +1,85 @@
...
+
+### GetAccessorStub
+
+```js
+interface GetAccessorStub <: Node {
+    type: "GetAccessorStub";
+}
+```
+
+A get accessor stub is a simplified version of grouped accessor. E.g. `class C { accessor c { get; } }`.
```
JLHwung, line 49:
```suggestion
A get accessor stub is used in auto-accessor. E.g. `class C { accessor c { get; } }`.
```

experimental/grouped-and-auto-accessors.md
```diff
@@ -0,0 +1,87 @@
...
+## ObjectAccessor
+
+```js
+interface ObjectAccessor <: Node {
+    type: "ObjectAccessor";
+    key: Literal | Identifier;
+    computed: boolean;
+    get: GetAccessorStub | GetAccessorMethod | null;
+    set: SetAccessorStub | SetAccessorMethod | null;
+}
```
JLHwung, line 40: In the latest commit I removed the inheritance from `Property`, because it will inherit `value: Expression` and `kind: "init" | "set" | "get"`. These properties are not application to the grouped accessor.

nzakas: Makes sense.

nzakas (approved):

PR #260, opened 2021-11-19
Add class brand check support

YuriTu:
[View Rendered Text](https://github.com/YuriTu/estree/blob/feat-brand-check/experimental/class-brand-check-expression.md)
---
This PR adds support on stage-1 [Proposal-class-brand-check](https://github.com/tc39/proposal-class-brand-check).

This PR is  very similar to [class property access expressions](https://github.com/estree/estree/blob/master/experimental/class-access-expressions.md), but `class brand check expressions` does not require a static members.

It seems that it can be used as part of `class property access expressions`.

However, the current `Identifier` is only `hasInstance`, maybe others will be added in the future, and it may still need to be separated in the future.


experimental/class-brand-check-expression.md
```diff
@@ -0,0 +1,19 @@
+# [class.hasInstance](https://github.com/tc39/proposal-class-brand-check)
+
+## ClassExpression
+
+```js
+extend interface ClassExpression {
+    operation: Identifier | null // operation.name currently only has : 'hasInstance'
```
JLHwung, line 7: I don't think we need to extend `ClassExpression`. For example, the RHS of 
```js
var c = class { }
```
is a class expression, but I don't think
```js
var c = class.hasInstance {}
```
will ever be valid.

In another hand, `class.hasInstance(instance)` is a call expression, if the callee `class.hasInstance` is a class expression, then its AST shape will share with `(class {})()`, though vastly different IMO.

j-f1: Prior art for this:
- `super.abcd()` uses a regular `MemberExpression` (and the `super()` syntax uses a regular `CallExpression`) with a special [`Super`](https://github.com/estree/estree/blob/master/es2015.md#expressions) node in the appropriate location
- `new.target` uses a [`MetaProperty`](https://github.com/estree/estree/blob/18bd9afbe8c348caca82057f1849614d98d6d0b6/es2015.md#metaproperty) with the identifier `new` as the value for the `meta` key.
- `import.meta` uses a [`MetaProperty`](https://github.com/estree/estree/blob/bae1f8084bfa1c45e31d08eb22ef6606e9697a07/es2020.md#metaproperty) with the identifier `import` as the value for the `meta` key.

YuriTu: There are some errors in my understanding, indeed `CallExpression` is more suitable. I will correct it later.

experimental/class-brand-check-expression.md
```diff
@@ -0,0 +1,28 @@
+# [class.hasInstance](https://github.com/tc39/proposal-class-brand-check)
+
+## ClassExpression
+
+```js
+interface Class <: Node {
+    type: "Class"
+}
+```
+A class pseudo-expression.
```
JLHwung, line 10: We already have https://github.com/estree/estree/blob/master/es2015.md#classes, that's why https://github.com/estree/estree/blob/master/experimental/class-access-expressions.md#class-property-access-expressions introduced `ClassObject`. I think we could share the design here.

experimental/class-brand-check-expression.md
```diff
@@ -0,0 +1,28 @@
...
+
+```js
+extend interface CallExpression {
+    callee: Class // currently only "class"
+    arguments: [ Expression ]
+}
+```
```
JLHwung, line 17: Some quick questions since the current proposal does not have a syntax spec.

Is
```js
class.hasInstance?.()
```
valid? If not we should add that `optional` must be `false`.

Is
```js
const hasInstance = class.hasInstance;
hasInstance(instance)
```
valid? If so we can't extend `CallExpression` because `class.hasInstance` itself can form an expression, just like `import.meta` and `function.sent` and thus not limited to be a `callee` of `CallExpression`.

experimental/class-brand-check-expression.md
YuriTu: I agree, `ClassObject` is more suitable.

experimental/class-brand-check-expression.md
YuriTu: There is a gay who has submitted a [spec](https://github.com/tc39/proposal-class-brand-check/pull/3) about the proposal.
But it has not been merged.
I think I need to confirm with the mantiner of the proposal whether it needs to support ChainExpression.

JLHwung: For reviewers: see https://github.com/tc39/proposal-class-brand-check/pull/3/files#diff-181371b08d71216599b0acccbaabd03c306da6de142ea6275c2135810999805aR173 for the proposed syntax.

```
CallExpression[Yield, Await] :
        CoverCallExpressionAndAsyncArrowHead[?Yield, ?Await] #callcover
        SuperCall[?Yield, ?Await]
        ImportCall[?Yield, ?Await]
        <ins>BrandCheckCall[?Yield, ?Await]</ins>
        CallExpression[?Yield, ?Await] Arguments[?Yield, ?Await]
        CallExpression[?Yield, ?Await] `[` Expression[+In, ?Yield, ?Await] `]`
        CallExpression[?Yield, ?Await] `.` IdentifierName
        CallExpression[?Yield, ?Await] TemplateLiteral[?Yield, ?Await, +Tagged]
        CallExpression[?Yield, ?Await] `.` PrivateIdentifier

      <ins>BrandCheckCall[Yield, Await] :</ins>
        `class` `.` `hasInstance` `(` AssignmentExpression[+In, ?Yield, ?Await] `)`
```

hax: The question is whether we want `class.hasInstance` be a real function/method. For example, `eval()` is a function but `import()` is not. As the author of the proposal, I prefer `import()` way because it's much simpler. If it was a method/function, we need to specify many behavior (such as whether it has receiver, aka. bound method or unbound), though we can specify the semantic anyway, it's hard to match everyone's expectation (a good/bad example is the trouble of `eval?.()` semantic --- indirect or direct eval). It seems make `class.hasInstance` be a real function/method do not have any real benefit, just introduce troubles.

YuriTu:
I discussed with the author of the proposal and decided  it would be like [importExpression](https://github.com/estree/estree/blob/master/es2020.md#importexpression). so I plan to add a new type of expression: `classHasInstanceExpression `

experimental/class-brand-check-expression.md
```diff
@@ -0,0 +1,21 @@
+# [class.hasInstance](https://github.com/tc39/proposal-class-brand-check)
+
+## ClassExpression
+
+```js
+extend interface classHasInstanceExpression <: Expression {
+    type: "classHasInstanceExpression",
+    source: Expression
```
JLHwung, line 8:
```suggestion
    instance: Expression
```

The `ImportExpression` has `source: Expression` because `"some-library"` is the imported source in `import("some-library")`. But I find `source` in `classHasInstanceExpression` confusing.
Other suggestions are welcome.

forivall: `candidate` is my suggestion. `argument` is also possible, but the syntax suggests that additional arguments may be added, so `argument` would become ambiguous.

YuriTu: Maybe `instance` is more appropriate, I hope to express the 
following infomation: 
1. `class.hasInstance` is not a real funtion object, not support multiple arguments. 
2.  The user can understand the role of `class.hasInstance` like `instanceof`. I hope this way could reduce the cost of understanding.

experimental/class-brand-check-expression.md
```diff
@@ -0,0 +1,21 @@
+# [class.hasInstance](https://github.com/tc39/proposal-class-brand-check)
+
+## ClassExpression
```
JLHwung, line 3:
```suggestion
## ClassHasInstanceExpression
```

experimental/class-brand-check-expression.md
```diff
@@ -0,0 +1,21 @@
...
+
+```js
+extend interface ClassHasInstanceExpression <: Expression {
+    type: "ClassHasInstanceExpression",
+    instance: Expression
+}    
+```
+
+eg :
```
JLHwung, line 12:
```suggestion
Example:
```

JLHwung (approved):

nzakas (approved):
LGTM

PR #263, opened 2021-12-01
Add async keyword to methodDefinition 

Cyberhan123:


JLHwung:
We don't need to add `async` to `MethodDefinition`, which is Babel AST. In ESTree, an async method has a `value` of type `FunctionExpression` whose `async` is `true`.

See also https://astexplorer.net/#/gist/6fa54333ea68fd86898f3c1765bed4da/37b17ccafc1ad7d25fdcb3fb7bb202221ea084af

Since the current `es2017.md` is correct, I suggest closing this PR.

Cyberhan123:
yes,you'r right

PR #272, opened 2022-01-21
Fixes Typo (extends -> extend)

sebiniemann:
I came across this by chance during an automatic merging of the spec files.


JLHwung (approved):
Thanks!

nzakas (approved):

PR #273, opened 2022-02-09
Align decorators to 2021-12 spec

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/decorators/experimental/decorators.md)

This PR aligns the decorators experimental AST to the latest proposal, which is asking for stage 3 advancement in March. The AST design is based on Babel's [current implementation](https://github.com/babel/babel/pull/13827).

[Decorators]: https://github.com/tc39/proposal-decorators

### DIFF
--- a/experimental/decorators.md
+++ b/experimental/decorators.md
@@ -28,6 +28,14 @@ extend interface Class {
 }
 ```
 
+## ClassBody
+
+```js
+extend interface ClassBody {
+    body: [ MethodDefinition | PropertyDefinition | StaticBlock | AccessorProperty ];
+}
+```
+
 ## MethodDefinition
 ```js
 extend interface MethodDefinition {
--- a/experimental/decorators.md
+++ b/experimental/decorators.md
@@ -1,4 +1,4 @@
-# [Decorators](https://github.com/wycats/javascript-decorators)
+# [Decorators](https://github.com/tc39/proposal-decorators)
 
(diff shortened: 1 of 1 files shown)


experimental/decorators.md
```diff
@@ -9,18 +9,36 @@ interface Decorator <: Node {
 }
 ```
 
-## Used
+## AccessorProperty
+```js
+interface AccessorProperty <: Node {
```
JLHwung, line 14: On AccessorProperty, an alternative design is to add `accessor: boolean` to `PropertyDefinition`. Babel introduced a new type so that we can support https://github.com/tc39/proposal-grouped-and-auto-accessors based on the new AST node type. See also discussions [here](https://github.com/babel/babel/pull/13681).

nzakas: I think a new node type makes sense. 👍

nzakas (approved):

nzakas:
@RReverser can you take a look?

RReverser:
Weird, I thought this Github status thing would show up to others automatically & limit notifications.

image

Anyway, sorry, but as a Ukrainian, I'm busy with other stuff and won't be looking at code in the nearest time. Feel free to proceed without me on any blockers.

nzakas:
Oh damn. Take care @RReverser. We are all pulling for Ukraine.

PR #275, opened 2022-02-22
js->ts;remove hard break line; jsonc -> js+json

xp44mm:
\`\`\`js->\`\`\`ts;
remove hard break line; 
\`\`\`jsonc -> \`\`\`js+\`\`\`json

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,10 +6,9 @@ Meanwhile JavaScript is evolving. This site will serve as a community standard f
 
 # AST Descriptor Syntax
 
-The spec uses a custom syntax to describe its structures.  For example, at the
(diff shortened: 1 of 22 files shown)


nzakas:
Thanks for helping. I’m not sure it makes much sense to change JS to ts considering that these blocks do not contain typescript code, but rather a custom syntax used just in ESTree.

xp44mm:
the `interface` and `enum` is keyword in ts. This is a step in the right direction.

nzakas:
Thanks, we will stick with what we have.

xp44mm:
you welcome!

PR #278, opened 2022-02-26
CallExpression

xp44mm:
The interface `CallExpression` has 2 presentations in `es2015.md`, should it be merged?


nzakas (approved):
I believe so. @JLHwung agree?

PR #283, opened 2022-04-06
Add missing "local" field

dpifke:
The description for ImportSpecifier and ExportSpecifier both refer to
a field named "local", which does not appear in either of the
preceding type definitions.


dpifke:
Bah, never mind—I just saw the parent ModuleSpecifier class, which defines local.  Sorry for the noise.

PR #287, opened 2022-09-13
add import-reflection

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/import-reflection/experimental/import-reflection.md)

This PR adds [Import Reflection] support.

[Import Reflection]: https://github.com/tc39/proposal-import-reflection

### DIFF
--- a/experimental/import-reflection.md
+++ b/experimental/import-reflection.md
@@ -6,10 +6,10 @@
 
 ```js
 extend interface ImportDeclaration {
-    reflection: "module" | null
+    module: boolean;
 }
 ```
-When `reflection` is `module`, the `specifiers` must be a length-1 array including `ImportDefaultSpecifier`, and the `assertions` must be an empty array.
+When `module` is `true`, the `specifiers` must be a length-1 array including `ImportDefaultSpecifier`, and the `assertions` must be an empty array.
 
 ## Expressions
 
--- a/experimental/import-reflection.md
+++ b/experimental/import-reflection.md
@@ -1,0 +1,28 @@
+# [Import Reflection][proposal-import-reflection]
+
+## Imports
+
+### ImportDeclaration
+
+```js
+extend interface ImportDeclaration {
+    reflection: "module" | null
+}
+```
+When `reflection` is `module`, the `specifiers` must be a length-1 array including `ImportDefaultSpecifier`, and the `assertions` must be an empty array.
+
+## Expressions
+
+### ImportExpression
+
+```js
+extend interface ImportExpression {
+    attributes: Expression | null;
+}
+```
+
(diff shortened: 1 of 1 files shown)


experimental/import-reflection.md
```diff
@@ -0,0 +1,28 @@
+# [Import Reflection][proposal-import-reflection]
+
+## Imports
+
+### ImportDeclaration
+
+```js
+extend interface ImportDeclaration {
+    reflection: "module" | null
```
JLHwung, line 9: Put an enum here so that we can support https://github.com/tc39/proposal-asset-references in the future. See also https://github.com/tc39/proposal-import-reflection#dynamic-import.

bradzacher: currently type-aware imports use `importKind` to specify this.
I.e.

`import type foo from 'foo'` has `importKind: 'type'`
`import typeof foo from 'foo'` has `importKind: 'typeof'`

Should we align with these or introduce a new property entirely?
Do we foresee cases where you might do `import type module foo from 'foo'`?

JLHwung: I though about `importKind`, currently Babel specifies `importKind: "value" for non-type imports. If we put `importKind: "module"` here, we should either add `importKind: "value"` to vanilla import declarations. But since the addition here is not backward-compatible, we have to recognize `importKind: undefined | null`, too.

> 'import type module foo from 'foo'

I don't think it is valid. I have gone ahead and banned this production in Babel.

bradzacher: We could spec that `importKind: null | undefined` is the same as `importKind: 'value'` for backwards compat.

The asset reference spec currently is `asset Foo from "foo";` without `import` - would we use the `ImportDeclaration` node for it or wouldn't it live as its own, separate node?

JLHwung: > The asset reference spec currently is asset Foo from "foo"; without import - would we use the ImportDeclaration node for it or wouldn't it live as its own, separate node?

Yes, you are right. Though authors of import reflection suggest that proposal reuse the import reflection syntax:

```js
import asset x from "<specifier>";
await import("<specifier>", { reflect: "asset" });
```

Anyway both proposals are still early so we could change the AST if the syntax change.

Although `import type module foo from 'foo'` makes no sense, `import typeof module Foo from './foo.wasm'` potentially works. In this case `Foo` can be the type of webassembly modules. If we are going with `importKind: "module"`, we also rule out the possibility that `import typeof module` might be a thing.

@nzakas Any opinion?

nzakas (approved):

experimental/import-reflection.md
JLHwung: @nicolo-ribaudo Thanks for the insider info. I will update the AST accordingly.

nzakas (approved):
I think this is fine as a starting point.

PR #288, opened 2022-09-15
add explicit resource management

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/explicit-resource-management/experimental/explicit-resource-management.md)

This PR adds [Explicit Resource Management] support.

[Explicit Resource Management]: https://github.com/tc39/proposal-explicit-resource-management

### DIFF
--- a/experimental/explicit-resource-management.md
+++ b/experimental/explicit-resource-management.md
@@ -10,6 +10,6 @@ extend interface VariableDeclaration {
 }
 ```
 
-If `kind` is `"using"`, for every declarator `d` of `declarations`, `d.id` must be an Identifier. If the variable declaration is the `left` of a ForInStatement / ForOfStatement, `d.init` must be `null`, otherwise `d.init` must be an Expression.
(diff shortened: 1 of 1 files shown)


nzakas (approved):
LGTM

experimental/explicit-resource-management.md
```diff
@@ -0,0 +1,15 @@
...
+
+## VariableDeclaration
+
+```js
+extend interface VariableDeclaration {
+    kind: "using";
+}
+```
+
+If `kind` is `"using"`, for every declarator `d` of `declarations`, `d.id` must be an Identifier. If the variable declaration is the `left` of a ForInStatement / ForOfStatement, `d.init` must be `null`, otherwise `d.init` must be an Expression.
```
JLHwung, line 13:
```suggestion
If `kind` is `"using"`, for every declarator `d` of `declarations`, `d.id` must be an Identifier. If the variable declaration is the `left` of a ForOfStatement, `d.init` must be `null`, otherwise `d.init` must be an Expression.
```
ForInStatement does not allow using declaration.

nzakas (approved):

PR #290, opened 2022-09-20
add extractors

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/extractors/experimental/extractors.md)

This PR adds [Extractors] support.

[Extractors]: https://github.com/tc39/proposal-extractors

### DIFF
--- a/experimental/extractors.md
+++ b/experimental/extractors.md
@@ -9,7 +9,6 @@ interface QualifiedName <: Node {
     right: Identifier;
 }
 ```
-An Identifier (`foo`) or dotted-name (`foo.bar`).
 
 ## ObjectPattern
 
--- a/experimental/extractors.md
+++ b/experimental/extractors.md
@@ -1,0 +1,29 @@
+# [Extractors for ECMAScript][proposal-extractors]
+
+## QualifiedName
+
+```js
+interface QualifiedName <: Node {
+    type: "QualifiedName";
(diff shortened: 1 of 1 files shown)


experimental/extractors.md
```diff
@@ -0,0 +1,29 @@
...
+## QualifiedName
+
+```js
+interface QualifiedName <: Node {
+    type: "QualifiedName";
+    left: QualifiedName | Identifier;
+    right: Identifier;
+}
+```
+An Identifier (`foo`) or dotted-name (`foo.bar`).
```
nzakas, line 12: Does this mean `left` can also be undefined?

JLHwung: No, `left` must be either an `Identifier` (`foo`) or a qualified name (`foo.bar`). The current AST design is borrowed from Babel's `TSQualifiedName`.

nzakas: Oh sorry, this note confused me due to its location. I see below in ObjectPattern and ArrayPattern that we’ve just added QualifiedName in addition to Identifier. 

Can we remove this paragraph? It doesn’t seem to clarify what’s already otherwise on this page.

JLHwung, line 12:
```suggestion
```

nzakas (approved):
LGTM. Thanks!

PR #293, opened 2022-11-23
rename ModuleDeclaration mixin

JLHwung:
In this PR we rename the `ModuleDeclaration` mixin so that it won't collide with the new AST type introduced in the [Module Declarations](https://github.com/tc39/proposal-module-declarations) proposal.

This is an editorial change since the AST structures are not impacted. The name `ImportOrExportDeclaration` is tentative so suggestions are definitely welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ time of writing, 'es2015.md' contained a description of `Program` as seen below
 ```js
 extend interface Program {
     sourceType: "script" | "module";
-    body: [ Statement | ModuleDeclaration ];
+    body: [ Statement | ImportOrExportDeclaration ];
 }
 ```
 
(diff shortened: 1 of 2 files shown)


nzakas:
It’s not clear to me why we need this mixin at all. What purpose is it serving in the doc?

JLHwung:
The `ModuleDeclaration` mixin was introduced in https://github.com/estree/estree/pull/79 to address https://github.com/estree/estree/issues/78.

I am either good with preserving the mixin or removing it. Either way it is an editorial change.

nzakas:
Ah thanks. Looks like I also initially opted for this approach 😆 so might as well keep it.

PR #294, opened 2022-11-24
add module declarations

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/module-declarations/experimental/module-declarations.md)

This PR adds [Module Declarations] support. <s>It builds on https://github.com/estree/estree/pull/293, I will rebase this PR once #293 is merged.</s>

[Module Declarations]: https://github.com/tc39/proposal-module-declarations

/cc @nicolo-ribaudo


experimental/module-declarations.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Declarations
+
+### ModuleDeclarations
+
+```js
+interface ModuleDeclaration <: Declaration {
+    type: "ModuleDeclaration";
+    id: Identifier;
+    body: Program;
```
nzakas, line 11: Allowing `Program` nodes anywhere other than the root will cause chaos for ESLint rules because that's the assumption right now. Could we maybe created a `ModuleBody` node here instead?

bradzacher: For reference this feature is syntactically very similar to TS's namespace/module syntax, and that is exactly what we chose - a block node to enclose the body.

- [TSModuleDeclaration](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/ast-spec/src/declaration/TSModuleDeclaration/spec.ts)
- [TSModuleBlock](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/ast-spec/src/special/TSModuleBlock/spec.ts)

experimental/module-declarations.md
```diff
@@ -0,0 +1,68 @@
+# [JavaScript Module Declarations][proposal-module-declarations]
+
+## Module block
+
+```js
+interface ModuleBlock <: Node {
+    type: "ModuleBlock";
+    sourceType: "module";
```
JLHwung, line 8: I chose `ModuleBlock` here because it starts with `{` and end with `}`, similar to `BlockStatement` and `StaticBlock`, so I slightly prefer `ModuleBlock` over `ModuleBody`.

The `sourceType` property is a redundant property so that `ModuleBlock` share the structure with `Program`. It will be convenient for AST consumers: scope queries like `scope.getProgramParent().sourceType` just work.

nzakas (approved):
This works for ESLint. 👍 I agree `ModuleBlock` is probably more appropriate than `ModuleBody`.

PR #301, opened 2023-02-08
fix: Clarify the value of the bigint Property

### LINKED ISSUE #296: Clearify how `BigIntLiteral.bigint` should represent
> `bigint` property is the string representation of the `BigInt` value. It doesn't include the suffix `n`.

https://github.com/estree/estree/blob/master/es2020.md#bigintliteral

1. Should numeric separators removed? (`1_2n`: `'12'` or `'1_2'`?)
2. Should it covert to 10-based? (`0xffn`: `'0xff'` or `'255'`?)
3. Should it be lowercased? (`0XaAn`: `'0xaa'` or `'0XaA'`?)

nzakas:
Clarifies the expectations of the string in the `BigIntLiteral.bigint` property.

Fixes #296

### DIFF
--- a/es2020.md
+++ b/es2020.md
@@ -20,8 +20,7 @@ interface BigIntLiteral <: Literal {
 }
 ```
 
-- `bigint` property is the string representation of the `BigInt` value.
-  It doesn't include the suffix `n`.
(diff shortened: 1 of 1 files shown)


JLHwung (approved):

marr4621-glitch:
###

PR #302, opened 2023-04-03
Import attributes

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/import-attributes/experimental/import-attributes.md)

This PR adds [Import Attributes] support. It reflects on Babel's current approach https://github.com/babel/babel/pull/15536. The Import Assertions proposal is superseded by this proposal.

[Import Attributes]: https://github.com/tc39/proposal-import-attributes

/cc @nicolo-ribaudo

### DIFF
--- a/experimental/import-attributes.md
+++ b/experimental/import-attributes.md
@@ -1,0 +1,62 @@
+# [Import Attributes][proposal-import-attributes]
+
+## Imports
+
+### ImportDeclaration
+
+```js
+extend interface ImportDeclaration {
+    attributes: [ ImportAttribute ];
+}
+```
+
+The `attributes` is non-empty when import attributes present, e.g., `import foo from "./foo.json" with { type: "json" }`.
+
+Note: While the spec also supports `assert { type: "json" }`, it was considered legacy and thus not covered here. If you want to support the legacy syntax, please use [Import Assertions](../stage3/import-assertions.md) instead.
+
+### ImportAttribute
+
+```js
+interface ImportAttribute <: Node {
+    type: "ImportAttribute";
+    key: Identifier | Literal;
+    value: Literal;
+}
+```
+
+An import attribute is an object-like key value pair, e.g. `type: "json"` in `import foo from "./foo.json" with { type: "json" }`. The `value` must be a string literal, that said, `value.value` is always `string`-type. If `key` is a `Literal`, it must be a string literal.
+
+## Exports
+
+### ExportNamedDeclaration
+
+```js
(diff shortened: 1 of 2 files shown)


nzakas (approved):
LGTM.

experimental/import-attributes.md
```diff
@@ -0,0 +1,62 @@
...
+## Exports
+
+### ExportNamedDeclaration
+
+```js
+extend interface ExportNamedDeclaration {
+    attributes: [ ImportAttribute ];
+}
+```
+- `attributes` must be an empty array when `source` is `null`.
```
fisker, line 38: Hello, @JLHwung, what is the reason behind this requirement?

experimental/import-attributes.md
```diff
@@ -0,0 +1,62 @@
...
+
+### ExportAllDeclaration
+
+```js
+extend interface ExportAllDeclaration {
+    attributes: [ ImportAttribute ];
+}
+```
+- `attributes` must be an empty array when `source` is `null`.
```
fisker, line 47: Is it possible to `export * as foo` without source?

experimental/import-attributes.md
fisker: Another reason to remove the requirement is that `specifiers` is not required to be an empty array in `ImportDeclaration`.

JLHwung: I think we will only consider the optional array approach if 1) `export {foo} with {} from "x"` is allowed and 2) `export {foo} with {} from "x"`is semantically difference than `export {foo} from "x"`. 

The non-optional-array dated back to early days of ESTree. For example, even though both `new A` and `new A()` are valid, the `arguments` of a `NewExpression` is always an array, given that `new A` is sematically equivalent with `new A()`: ESTree is an AST spec, it does not cover differences from the CST perspective.

JLHwung: > `specifiers` is not required to be an empty array in `ImportDeclaration`.

Could you given some pointers? This does not seem to be aligned with what I read from the `es2015.md`:

https://github.com/estree/estree/blob/52399d923111fe2a4dcfe1ede795e08356f35ddc/es2015.md?plain=1#L330-L337

fisker: Sorry, "specifiers" is a bad example, please ignore.

> I think we will only consider the optional array approach

I get the point.

BTW, in your code examples, `from` should come before `with`, you may want edit them, just avoid confusion if others read this discussion.

Thanks for the explanation.

PR #306, opened 2023-05-17
Add AST definition for the Source Phase Imports proposal

nicolo-ribaudo:
The Import Reflection proposal (https://github.com/tc39/proposal-import-reflection/) has been updated to use the following syntax:
- `import source X from "X"`
- `import.source("X")`

It is now part of a bigger set of features, that allow importing different "phases" of a module:
- `import source X from "X"`
- `import instance X from "X"` (which is currently not tracked in any proposal, but it is being considered as part of the intersection between import reflection and https://github.com/tc39/proposal-compartments/blob/master/0-module-and-module-source.md)
- `import defer * as ns from "X"` - https://github.com/tc39/proposal-defer-import-eval/

You can read [this presentation](https://docs.google.com/presentation/d/1F62Jia5erIm6m6nqkm_2pFIlNLOVF0E4ewrVRytSJEs/edit#slide=id.p) for more details.

All these "phases" are not composable, so I chose to represent them in the AST as an enum instead of boolean properties. However, this PR only adds support for `source` since it's the only one introduced by this proposal.

I am on the fence regarding what to do with `import x from "X"`. Should `phase` be `null` since it is the default phase, or should we always set `phase` to a string and thus something like `"full"`/`"evaluation"`?

(PS. The ESTree AST works much better than the Babel AST when trying to define `import.source()` 😄)

### DIFF
--- a/experimental/import-source-phase.md
+++ b/experimental/import-source-phase.md
@@ -1,0 +1,29 @@
+# [Source Phase Imports][proposal-import-reflection]
+
+## Imports
+
+### ImportDeclaration
+
+```js
+extend interface ImportDeclaration {
+    phase: "source" | null;
+}
+```
+
+`phase` is `"source"` when representing an import in the form `import source X from "X"`.
+
+When `phase` is `"source"`, the `specifiers` must be a length-1 array including `ImportDefaultSpecifier`.
+
+## Expressions
+
+### ImportExpression
+
+```js
+extend interface ImportExpression {
+    phase: "source" | null;
+}
+```
+
+`phase` is `"source"` when representing a dynamic import in the form `import.source("X")`.
+
+[proposal-import-reflection]: https://github.com/tc39/proposal-import-reflection
--- a/experimental/source-phase-imports.md
+++ b/experimental/source-phase-imports.md
@@ -1,4 +1,4 @@
-# [Source Phase Imports][proposal-import-reflection]
+# [Source Phase Imports][proposal-source-phase-imports]
 
 ## Imports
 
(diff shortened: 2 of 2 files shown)


JLHwung:
Can you add a deprecation note on https://github.com/estree/estree/blob/master/experimental/import-reflection.md? We have done something similar in https://github.com/estree/estree/blob/master/stage3/import-assertions.md. Or we can just remove `import-reflection` since that syntax is no longer proposed.

experimental/import-source-phase.md
```diff
@@ -0,0 +1,29 @@
...
+
+```js
+extend interface ImportDeclaration {
+    phase: "source" | null;
+}
+```
+
+`phase` is `"source"` when representing an import in the form `import source X from "X"`.
+
+When `phase` is `"source"`, the `specifiers` must be a length-1 array including `ImportDefaultSpecifier`.
```
JLHwung, line 15: Should we also add that "if `attributes` is defined, it must be an empty array"? Since `import source X from "foo" with { type: "json" }` is not a thing, right?

JLHwung:
The proposal's url slash `import-reflection` is not renamed to `import-source-phase`, in this case maybe we can just replace the current `import-reflection` docs with the new one.

nzakas:
@nicolo-ribaudo are you planning on updating this PR based on that last comment?

nicolo-ribaudo:
@guybedford @lcasdev Any plans to rename the proposal repo given the name change?

nicolo-ribaudo:
PR updated to match the new proposal URL: https://github.com/tc39/proposal-source-phase-imports

JLHwung (approved):

nzakas (approved):
LGTM.

PR #307, opened 2023-06-06
rename assertions/attributes in ImportExpression

### LINKED ISSUE #305: Rename `assertions`/`attributes` in `ImportExpression`
In `import(src, { assert: { type: "json" }, somethingElse: true })`, only `{ type: "json" }` is the assertions bag. However, ESTree AST for the above code calls `{ assert: { type: "json" }, somethingElse: true }` "assertions". The same happens with import attributes, where only the object stored in the `with` property is the attributes part.

It's probably too late to change the ESTree shape for import assertions, but maybe we could change it for import attributes by renaming `assertions` to a more generic `options`?

```ts
extend interface ImportExpression {
    options: Expression | null;
}
```

JLHwung:
Resolves #305.

### DIFF
--- a/experimental/import-attributes.md
+++ b/experimental/import-attributes.md
@@ -52,11 +52,11 @@ extend interface ExportAllDeclaration {
 
 ```js
 extend interface ImportExpression {
-    attributes: Expression | null;
+    options: Expression | null;
 }
 ```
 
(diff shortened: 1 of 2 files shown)


nzakas (approved):
LGTM

PR #308, opened 2023-07-20
es5: Correct Property description

bojavou:
The description for `Property` says the value may be either a string or a number.

>A literal property in an object expression can have either a string or number as its `value`.

I think this meant to say `key` instead of `value`. Any expression may be used as the value. The key can be a string or number.

```js
{ 1: 2 + 2 } // { '1': 4 }
{ 'key': 2 + 2 } // { 'key': 4 }
```

This patch updates the description of `Property` to constrain `key` instead of `value`.

### DIFF
--- a/es5.md
+++ b/es5.md
@@ -510,7 +510,7 @@ interface Property <: Node {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


nzakas (approved):
LGTM. Thanks!

PR #309, opened 2023-07-21
Align ImportAttribute.key constraint

bojavou:
The description of `ImportAttribute` constrains a `key` to strings only.

>If `key` is a `Literal`, it must be a string literal.

The latest spec also allows numeric literals, as are possible in an object literal.

```js
import gadget from 'gadget' with {
  1: 'value1',
  2n: 'value2'
}
```

https://ci.tc39.es/preview/tc39/ecma262/sha/1284bb51257753a50d9b29056486da245e38ac03/#prod-WithEntries

```
WithEntries :
        LiteralPropertyName : StringLiteral
        LiteralPropertyName : StringLiteral , WithEntries
```

```
LiteralPropertyName :
        IdentifierName
        StringLiteral
        NumericLiteral
```

This patch extends the constraint on `ImportAttribute.key` to allow numeric literals.

### DIFF
--- a/experimental/import-attributes.md
+++ b/experimental/import-attributes.md
@@ -24,7 +24,7 @@ interface ImportAttribute <: Node {
 }
 ```
 
-An import attribute is an object-like key value pair, e.g. `type: "json"` in `import foo from "./foo.json" with { type: "json" }`. The `value` must be a string literal, that said, `value.value` is always `string`-type. If `key` is a `Literal`, it must be a string literal.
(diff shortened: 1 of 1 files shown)


JLHwung:
It was relaxed to `LiteralPropertyName` in the June 2023 stage 3 review of this proposal: https://github.com/tc39/ecma262/pull/3057#discussion_r1228752641

There is an ongoing discussion about restricting it to string / identifier again: https://github.com/tc39/proposal-import-attributes/issues/145

As of July 2023, Babel has not implemented support for numerical attribute key yet. I think we should wait until the upstream spec settles down on the allowed syntax here. After all, using a numeric literal here is pretty rare.

bojavou:
Shoot, I just implemented numeric literals in a parser. Thought it seemed strange when I read the spec. Following that thread now too so I can stay aligned.

PR #318, opened 2024-11-11
add discard-binding

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/discard-binding/experimental/discard-binding.md)

This PR adds [`void` Discard Bindings for ECMAScript] support.

Although `const [, x] = foo` is equivalent to `const [void, x] = foo`. I chose a dedicate AST node `Void` instead of `null` because 1) they have different AST representation, which is important for Babel's feature detection and 2) comments around the `void` token can be attached to this AST node. 

An alternative design will be to avoid extending from `Pattern` but instead extend from the element types of the function parameters, array patterns, object patterns and variable declarators. This will be mostly editorial as the AST shape for end users are not changed, but we can get rid of the clauses where `Void` can not be used. I slightly prefer the current version as `void` is also descended from the Patter production in the proposal spec text.

/cc @Boshen, @JoshuaKGoldberg and @kdy1 since you also maintained ESTree-compatible parsers. Please share if you have any ideas about the AST design.

[`void` Discard Bindings for ECMAScript]: https://github.com/tc39/proposal-discard-binding

### DIFF
--- a/experimental/discard-binding.md
+++ b/experimental/discard-binding.md
@@ -2,14 +2,14 @@
 
 # Patterns
 ```js
-interface Void <: Pattern {
-    type: "Void";
+interface VoidPattern <: Pattern {
+    type: "VoidPattern";
 }
 ```
 
-- The `left` of an `AssignmentPattern` must not be `Void`
-- The `argument` of a `RestElement` must not be `Void`
-- If the `id` of a `VariableDeclarator` is `Void`, the kind of the encompassing `VariableDeclaration` must be either `using` or `await using` defined in the [Explicit Resource Managemenr Proposal].
+- The `left` of an `AssignmentPattern` must not be `VoidPattern`
+- The `argument` of a `RestElement` must not be `VoidPattern`
+- If the `id` of a `VariableDeclarator` is `VoidPattern`, the `kind` of the encompassing `VariableDeclaration` must be either `using` or `await using` defined in the [Explicit Resource Management Proposal].
 
 [Explicit Resource Management Proposal]: ../stage3/explicit-resource-management.md
 [proposal-discard-biding]: https://github.com/tc39/proposal-discard-binding
--- a/experimental/discard-binding.md
+++ b/experimental/discard-binding.md
@@ -11,5 +11,5 @@ interface Void <: Pattern {
 - The `argument` of a `RestElement` must not be `Void`
 - If the `id` of a `VariableDeclarator` is `Void`, the kind of the encompassing `VariableDeclaration` must be either `using` or `await using` defined in the [Explicit Resource Managemenr Proposal].
 
-[Explicit Resource Managemenr Proposal]: ../stage3/explicit-resource-management.md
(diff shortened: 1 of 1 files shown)


kdy1 (changes requested):
Stage 1 proposals should never be added to estree

JLHwung:
To clarify, this PR does not change the ESTree spec. Instead, any early proposals (stage 0 - 2) are first added to the `experimental` folder first, and if the proposal is advanced to stage 3, they will be moved to the `stage3` folder. If the proposal is advanced to stage 4, they will be merged to estree. In most circumstances, they will be merged directly without further changes.

In the past few years we have been following this process: e.g. https://github.com/estree/estree/pull/250 moves class features to ES2022.

My intention is to engage community stakeholders in the AST design of early proposals, and avoid AST design fragmentations because it is expensive for tools to convert from one flavour to another flavour just because they are slightly different in different tools.

bradzacher:
@JLHwung is correct -- the point of the `experimental` folder is to allow declaration of early-stage features to allow tools to collaborate and standardise on a single AST spec ahead of the feature's release.

This avoid situations like with class properties or optional chaining where much of the ecosystem went in one direction, but the final ESTree AST went in a different direction, causing a fracture in the ecosystem that took significant effort to fix.

bradzacher (approved):

Boshen (approved):

kdy1 (approved):

experimental/discard-binding.md
```diff
@@ -0,0 +1,15 @@
...
+interface Void <: Pattern {
+    type: "Void";
+}
+```
+
+- The `left` of an `AssignmentPattern` must not be `Void`
+- The `argument` of a `RestElement` must not be `Void`
+- If the `id` of a `VariableDeclarator` is `Void`, the kind of the encompassing `VariableDeclaration` must be either `using` or `await using` defined in the [Explicit Resource Managemenr Proposal].
+
+[Explicit Resource Managemenr Proposal]: ../stage3/explicit-resource-management.md
```
nzakas, line 14:
```suggestion
[Explicit Resource Management Proposal]: ../stage3/explicit-resource-management.md
```

experimental/discard-binding.md
```diff
@@ -0,0 +1,15 @@
+# [`void` Discard Bindings for ECMAScript](proposal-discard-binding)
+
+# Patterns
+```js
+interface Void <: Pattern {
```
nzakas, line 5: My only concern is that this name could easily be confused as representing any `void` keyword.

What about calling it `VoidPattern` for clarity instead?

JLHwung: I am good with that. Previously `Void` was picked because we also have a `Super` node, but since the `super` keyword can only serve for that purpose, `VoidPattern` sounds good to me.

JoshuaKGoldberg (approved):
Thanks for including us in review - this looks good to me! +1 to calling it the more precise `VoidPattern`.

PR #319, opened 2024-11-13
add structs

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/structs/experimental/structs.md)

This PR supports the stage 2 proposal [JavaScript Structs]. Similar to #318, this PR only touches the `experimental` folder, so the ESTree spec is not changed.

/cc @Boshen, @bradzacher, @JoshuaKGoldberg and @kdy1.

[JavaScript Structs]: https://github.com/tc39/proposal-structs

### DIFF
--- a/experimental/structs.md
+++ b/experimental/structs.md
@@ -1,12 +1,27 @@
 # [JavaScript Structs: Fixed Layout Objects and Some Synchronization Primitives](proposal-structs)
 
+# Classes
+
+```js
+interface ClassOrStruct <: Node {
+    id: Identifier | null;
+    superClass: Expression | null;
+}
+```
+
+```js
+extend interface Class <: ClassOrStruct {
+    body: ClassBody
+}
+```
+
+Note: The `Class` interface changes are editorial. This proposal does not incur any AST changes to `ClassDeclaration` and `ClassExpression`.
+
 # Structs
 
 ```js
-interface Struct <: Node {
+interface Struct <: ClassOrStruct {
     shared: boolean;
-    id: Identifier | null;
-    superClass: Expression | null;
     body: StructBody;
 }
 ```
--- a/experimental/structs.md
+++ b/experimental/structs.md
@@ -1,13 +1,30 @@
 # [JavaScript Structs: Fixed Layout Objects and Some Synchronization Primitives](proposal-structs)
 
-# Declarations
+# Structs
+
+```js
+interface Struct <: Node {
+    shared: boolean;
+    id: Identifier | null;
+    superClass: Expression | null;
+    body: StructBody;
+}
+```
 
 ## StructDeclaration
 ```js
-interface StructDeclaration <: Class, Declaration {
+interface StructDeclaration <: Struct, Declaration {
     type: "StructDeclaration";
     id: Identifier;
-    shared: boolean;
+}
+```
+
+## StructBody
+
+```js
+interface StructBody <: Node {
+    type: "StructBody";
+    body: [ MethodDefinition | PropertyDefinition | StaticBlock ]
 }
 ```
 
@@ -27,7 +44,7 @@ interface UnsafeBlockStatement <: BlockStatement {
 
 ### ExportDefaultDeclaration
 ```js
-interface AnonymousDefaultExportedStructDeclaration <: Class {
+interface AnonymousDefaultExportedStructDeclaration <: Struct {
     type: "StructDeclaration";
     id: null;
 }
--- a/experimental/structs.md
+++ b/experimental/structs.md
@@ -1,0 +1,40 @@
+# [JavaScript Structs: Fixed Layout Objects and Some Synchronization Primitives](proposal-structs)
+
+# Declarations
+
+## StructDeclaration
+```js
+interface StructDeclaration <: Class, Declaration {
+    type: "StructDeclaration";
+    id: Identifier;
+    shared: boolean;
+}
+```
+
+# Statements
+
+## UnsafeBlockStatement
+
+```js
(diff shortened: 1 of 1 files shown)


experimental/structs.md
```diff
@@ -0,0 +1,57 @@
...
+}
+```
+
+# Statements
+
+## UnsafeBlockStatement
+
+```js
+interface UnsafeBlockStatement <: BlockStatement {
+    type: "UnsafeBlockStatement";
```
JLHwung, line 37: An alternative design is to add `shared: boolean` to the `BlockStatement` node. However, in ESTree the `BlockStatement` node is also used in FunctionDeclaration, I think it is not worthy to add a boolean property to such a popular AST node when we can confirm it is always `false` in this situation.

RReverser: > An alternative design is to add `shared: boolean`

I assume you meant `unsafe: boolean` since "shared" applies to structs not blocks?

I agree a separate type makes sense, although it's slightly unfortunate that traverse algorithms will have to add a new node type even though it should be treated 1:1 as any other blocks.

experimental/structs.md
```diff
@@ -0,0 +1,57 @@
+# [JavaScript Structs: Fixed Layout Objects and Some Synchronization Primitives](proposal-structs)
+
+# Structs
+
+```js
+interface Struct <: Node {
+    shared: boolean;
+    id: Identifier | null;
```
RReverser, line 8: As far as I can tell from the proposal, `struct` must always have a name so `| null` is not valid here.

RReverser: Basically, there should be just `StructDeclaration` and you can probably reuse `<: ClassDeclaration` for it as a base.

experimental/structs.md
```diff
@@ -0,0 +1,57 @@
+# [JavaScript Structs: Fixed Layout Objects and Some Synchronization Primitives](proposal-structs)
+
+# Structs
+
+```js
+interface Struct <: Node {
```
RReverser, line 6: Should we reuse `<: Class` since they're supposed to be very similar?

experimental/structs.md
JLHwung: Yes, I was meant for `unsafe: boolean`.
> even though it should be treated 1:1 as any other blocks.
Right, we already have `StaticBlock` as a precedent case, which also extends from `BlockStatement`.

An alternative design is to model `UnsafeBlockStatement` as a wrapper of `BlockStatement`:
```js
interface UnsafeBlockStatement <: Statement {
  body: BlockStatement
}
```
The Pros of this design is that linters will probably don't have to specifically handle `UnsafeBlockStatement`. 
The cons is the extra hierarchy.

bradzacher: I think a new node is better because it means we  can easily limit any future nodes that might only be allowed in unsafe blocks.

bradzacher (approved):

experimental/structs.md
```diff
@@ -0,0 +1,55 @@
+# [JavaScript Structs: Fixed Layout Objects and Some Synchronization Primitives](proposal-structs)
+
+# Structs
+
+```js
+interface Struct <: Class {
```
bradzacher, line 6: Is it correct to extend from class? I'm not sure it is considering shared structs are limited in what they allow (eg no private members)

bradzacher: Quoting the readme for shared structs:

> They cannot contain instance methods or instance private names.

Non-shared structs just say they share their syntax with classes. I haven't had a chance to read the spec itself though.

JoshuaKGoldberg: I believe Brad is right. As I understand it, Struct is more of a _sibling_ to Class than a _descendent_ - because of Brad's quoted note on non-shared syntax.

As a novice to TC39 proposals, I found the AST naming a bit confusing in that the general "Struct" is less restricted than the more specific "SharedStruct". I found it easier to think of them as named "UnsharedStruct" and "SharedStruct" per the spec text.

Putting it together, the way I'm understanding the hierarchy is roughly:

* `Class` and `SharedStruct` extend `Declaration`
* (`Unshared`)`Struct` extends `Class`, `SharedStruct`

Is that roughly accurate?

JLHwung: > They cannot contain instance methods or instance private names.

Yes, a share construct can not contain non-constructor instance methods/instance private names, however, they could contain static methods/static private names ([spec](https://tc39.es/proposal-structs/#sec-struct-definitions-static-semantics-containsinstancemethod)), which are modelled as MethodDefinition/PropertyDefinition with `static: true`. Although the shared struct is more limited than the non-shared struct, they do share the same element types. So I added the `shared: boolean` to model both of them.

Note that whether the interface `Struct` extends from `Class` is mostly editorial because they are not the final AST shape: They are intermediate interface that helps us structure the ESTree spec. If we agree that `Struct` should be considered a sibling of `Class`, I can revert the last commit.

RReverser: >  If we agree that Struct should be considered a sibling of Class, I can revert the last commit.

Yeah if we do that I'd just use a base interface mostly to ensure that we don't accidentally diverge in property names between the structs and classes.

JLHwung: Okay, I should confirm the editorial convention here: If we extract a new base interface, let's say `ClassOrStruct` (better names are welcomed), should I directly change the `es2015.md` in this PR? Or should I put those changes in the `structs.md`?

RReverser: Just in `structs.md`, yeah. So you'd have something like `extend interface Class <: ClassOrStruct {}` followed by declaration for struct itself.

RReverser (approved):

PR #327, opened 2025-06-23
add deferred-reexports

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/deferred-reexports/experimental/deferred-reexports.md)

This PR adds [Deferred re-exports] support.

[Deferred re-exports]: [proposal-url](https://tc39.es/proposal-deferred-reexports/#sec-exports)

### DIFF
--- a/experimental/deferred-reexports.md
+++ b/experimental/deferred-reexports.md
@@ -1,0 +1,27 @@
+# [Deferred re-exports][proposal-deferred-reexports]
+
+## Exports
+
+### ExportNamedDeclaration
+
+```js
+extend interface ExportNamedDeclaration {
+    phase: "defer" | null;
+}
+```
+
+`phase` is `"defer"` when representing an export in the form `export defer { foo } from "X"`.
+
+When `phase` is `"defer"`, the `declaration` must be `null`, the `source` must be a `Literal`.
+
+### ExportAllDeclaration
+
+```js
+extend interface ExportAllDeclaration {
+    phase: "defer" | null;
+}
+```
+
+`phase` is `"defer"` when representing an export in the form `export defer * from "X"`.
+
+[proposal-deferred-reexports]: https://github.com/tc39/proposal-deferred-reexports


experimental/deferred-reexports.md
```diff
@@ -0,0 +1,27 @@
+# [Deferred re-exports][proposal-deferred-reexports]
+
+## Exports
+
+### ExportNamedDeclaration
+
+```js
+extend interface ExportNamedDeclaration {
+    phase: "defer" | null;
```
nzakas, line 9: I'm curious about the name "phase" here. Are there other potential phases?

Any reason not to use `defer: boolean`?

nicolo-ribaudo: For _imports_, if it wasn't for top-level await then `defer` would truly be a "phase":
1. loading of a module (i.e. `import source`)
2. loading of a module and its dependencies (i.e. `import defer`)
3. evaluation of a module and its dependencies (i.e. just `import`)

top-level await makes the distinction between 2 and 3 a bit blurry because `import defer` actually does _some_ evaluation.

---

When it comes to `export ... from`, the idea is that `export defer` is _similar_ to `import defer`, however where it lives in the phases is a bit more blurry. It does two things:
- in (2) and (3), completely skip those phases for the unused modules
- like `import defer`, when used together with a `import *` it "pauses" the process for the `export defer`-ed module at phase (2) deferring phase (3).

nzakas: Thanks for the background. I think my overall question is if `defer` can be coupled with anything else? Is there a chance you'd ever want to do `import defer source`?

bradzacher: Does it make sense to do the same thing that everyone does for `import type`?

Right now the `kind` property is set to either `type` or `value` depending on the presence or absence of the keyword. 

Based on TS 5.9's implementation of `import defer`, you cannot do `import defer type` or `import type defer`.

Do we envision that `import type source` would be a thing? If not then aligning with the ecosystem on `kind` might be a better option.

bradzacher: I guess with the import reexport already being landed we probably won't align?
https://github.com/estree/estree/blob/master/stage3/defer-import-eval.md?plain=1#L9

Would be good to not have two properties that are specifying a keyword when those keywords are conflicting / can't be used at the same time.

PR #334, opened 2026-01-20
add partial-application

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/partial-application/experimental/partial-application.md)

This PR adds [Partial Application] support. It is based on Babel's current approach, which implements a legacy version of this proposal prior to October 2021.

[Partial Application]: https://github.com/tc39/proposal-partial-application

### DIFF
--- a/experimental/partial-application.md
+++ b/experimental/partial-application.md
@@ -26,7 +26,7 @@ interface PartialCallExpression <: ChainElement {
 ## PartialNewExpression
 
 ```js
-interface PartialNewExpression <: ChainElement {
+interface PartialNewExpression <: Node {
     type: "PartialNewExpression";
     callee: Expression;
     arguments: [ Expression | SpreadElement | ArgumentPlaceholder | RestPlaceholder ];
--- a/experimental/partial-application.md
+++ b/experimental/partial-application.md
@@ -1,0 +1,29 @@
+# [Partial Application](proposal-partial-application)
+
+# Expressions
+
+```js
+interface PartialElement <: Node {
+    partial: boolean;
+    arguments: [ Expression | SpreadElement | ArgumentPlaceholder | RestPlaceholder ];
+}
+
+interface ArgumentPlaceholder <: Node {
+    type: "ArgumentPlaceholder";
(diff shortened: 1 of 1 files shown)


michaelficarra:
I don't like this. Partial applications are not like variants of call/new expressions. They're unrelated other than they have similar surface syntax. Partial call/new should be their own nodes.

JLHwung:
> I don't like this. Partial applications are not like variants of call/new expressions. They're unrelated other than they have similar surface syntax. Partial call/new should be their own nodes.

Thank you for your suggestion. Like I said the current AST design reflects on Babel's current approach, implemented long ago when the proposal only contains the argument placeholder. Another reason for `partial: boolean` is that ESTree models `?.()` as

```
ChainExpression > CallExpression { optional: true }
```

I am fine with new nodes because of the `~` prefix requires certain marker on the call/new expression. Either a boolean property or a new node type can serve for such purpose. I will revise the AST draft.

Aside: As a known deviation from ESTree, Babel models `?.()` as an `OptionalCallExpression` node, so I am looking for naming suggestions for the `?.~()` production, do you think `OptionalPartialCallExpression` is fine? or `PartialOptionalCallExpression`?

JLHwung:
@michaelficarra Updated the AST draft, PTAL.

experimental/partial-application.md
```diff
@@ -0,0 +1,41 @@
+# [Partial Application](proposal-partial-application)
+
+# Expressions
+
+```js
+interface PartialElement <: Node {
+    partial: boolean;
+    arguments: [ Expression | SpreadElement | ArgumentPlaceholder | RestPlaceholder ];
+}
```
michaelficarra: I think this is unused now.

michaelficarra:
Yes this seems better. Thank you.

> I am looking for naming suggestions

I have no opinions on naming.

JLHwung:
@nzakas ping. Any review comments? Thank you.

nzakas (approved):
I like the revised approach. LGTM.

PR #335, opened 2026-02-25
add import-sync

JLHwung:
## [View Rendered Text](https://github.com/JLHwung/estree/blob/import-sync/experimental/import-sync.md)

This PR adds [Import Sync](https://github.com/tc39/proposal-import-sync) support.

### DIFF
--- a/experimental/import-sync.md
+++ b/experimental/import-sync.md
@@ -6,10 +6,8 @@
 
 ```js
 extend interface ImportExpression {
-    phase: "sync" | null;
+    sync: boolean;
 }
 ```
-
-`phase` is `"sync"` when representing a dynamic import in the form `import.sync("X")`.
 
 [proposal-import-sync]: https://github.com/tc39/proposal-import-sync
--- a/experimental/import-sync.md
+++ b/experimental/import-sync.md
@@ -1,0 +1,15 @@
+# [Import Sync][proposal-import-sync]
+
+## Expressions
+
+### ImportExpression
+
+```js
+extend interface ImportExpression {
+    phase: "sync" | null;
+}
+```
+
+`phase` is `"sync"` when representing a dynamic import in the form `import.sync("X")`.
+
+[proposal-import-sync]: https://github.com/tc39/proposal-import-sync


experimental/import-sync.md
```diff
@@ -0,0 +1,15 @@
+# [Import Sync][proposal-import-sync]
+
+## Expressions
+
+### ImportExpression
+
+```js
+extend interface ImportExpression {
+    phase: "sync" | null;
```
JLHwung: The proposal explicitly states that [`sync` is not a `phase`](https://github.com/tc39/proposal-import-sync/blob/main/README.md#is-import-sync-a-module-loading-phase), but it is very convenient for us to let it stay in the `phase` enum because of the proposed `import.sync` syntax.

It is unclear how this proposal should interact with other import phases proposal (https://github.com/tc39/proposal-import-sync/issues/12), but we are sure that `source` and `sync` must be mutual-exclusive after `import.`, if in the future `import.sourceSync` were introduced, we can still expand the `phase` enum.

An alternative design is to add `sync: boolean` to `ImportExpression`. Although this approach allows the integration between `source` and `sync`, it does complicates the code generator part.

/cc @nicolo-ribaudo

nzakas: > An alternative design is to add `sync: boolean` to `ImportExpression`. Although this approach allows the integration between `source` and `sync`, it does complicates the code generator part.

Can you explain a bit about how the code generator is affected?

nicolo-ribaudo: The proposal at the moment doesn't define something like "a sync import of a source" or "a sync import of a deferred import", but I think we should keep the AST future-proof to it as there is no technical reason why those two should not be eventually supported by JS.

JLHwung: I think both approach both approach is future-proof, as they can be extended to support the "sync import of source" feature:

---

In the current approach, the `phase` of `ImportExpression` is essentially any identifier name after `import.`, so if there is "sync import of source":

```js
extend {
  phase: "sourceSync" // or "syncSource" if proposed the other way
}
```
No changes are required for the code generator.

---

In the alternative approach: no AST change is required as `{ phase: "source", sync: true }` can model either `import.sourceSync` or `import.syncSource`. But the code generator may have to uppercase `phase` or print `Sync` after `phase` is printed.

michaelficarra: @JLHwung The same logic can be used to stuff every node into a single node type. That doesn't make it a good design. We should not be optimising the AST for the code generator's convenience to just plop one of the values directly into the output.

nzakas: I think the `{ phase: "source", sync: true }` approach fits better into how we've typically done things in ESTree. All of the necessary data is there to generate the correct syntax.

JLHwung: OK, revised to the boolean approach.

nzakas (approved):
LGTM.

PR #340, opened 2026-08-07
[deferred-reexports] Add `import { x, y } as ns` support

nicolo-ribaudo:
This is adding AST for this syntax:
```js
import { x, y } as ns from "foo";
```

introduced by https://tc39.es/proposal-deferred-reexports/#sec-imports. More info is in the TC39 presentation that introduced the syntax: https://docs.google.com/presentation/d/1uYgMUqz5bZT3wwPARnyq7GrJlWFesIrmoCJA1iDH3W8/edit?slide=id.g3965d76ffd6_0_22#slide=id.g3965d76ffd6_0_22

I also need to add support for `export { x, y } as ns from "foo"`, however depending on the outcome of https://github.com/estree/estree/issues/339 I see two possible options. If we reuse `ImportNamespaceSpecifier` for `export * as ns from "x"` then there is nothing to do, otherwise if the `ExportAllDeclaration` gets a `.specifier: Identifier | Literal` field, I'd also add it to `ExportNamedDeclaration` for this proposal.


experimental/deferred-reexports.md
```diff
@@ -1,5 +1,18 @@
 # [Deferred re-exports][proposal-deferred-reexports]
 
+## Imports
+
+### ImportNamespaceSpecifier
+
+```js
+extend interface ImportNamespaceSpecifier <: ModuleSpecifier {
+    exportsFilter: [ Identifier | Literal ] | null;
```
nicolo-ribaudo, line 9: For the name here I'd align with whatever the outcome of https://github.com/tc39/proposal-deferred-reexports/issues/39 is (and you are welcome to share opinions :) )

experimental/deferred-reexports.md
JLHwung, line 38:
```suggestion
`phase` is `"defer"` when representing an export in the form `export defer * as ns from "X"`.
```

experimental/deferred-reexports.md
JLHwung: The `exportsFilter`, or simply `filter`, both look good to me.

Will `import { foo as bar } as ns from "./lib"` ever be a thing? If so I suggest we use `ImportSpecifier` for `exportsFilter`.

The current proposed structure serves the current proposal well, but if there is ever a proposal extending the current syntax, there is not much room to nicely extend.

